PR #24, opened 2026-03-03
docs: update README

exo404:


daopunk (approved):
approved.
in the future we could organize by operating system with nested READMEs, so a dev could click their OS to get specific instructions. and keep nested .md docs in a folder at root.
for example: https://github.com/defi-wonderland/aztec-standards/blob/dev/README.md
(example is with nested token READMEs)

PR #46, opened 2026-06-27
ci: add typecheck, lint, and unit test quality gate

logicminds:
Adds a CI quality job that runs `pnpm check`, `pnpm lint`, and `pnpm test` on push to main and on pull requests.

All three steps use `continue-on-error: true` for now so existing diagnostics do not block merges while they are being addressed.

Changes:
- New `.github/workflows/ci.yaml` quality job
- New `eslint.config.js` with TypeScript + Svelte flat config
- New `lint` npm script and ESLint dev dependencies

Verification:
- `pnpm check` runs
- `pnpm lint` runs
- `pnpm test` passes (290 tests in 57 files)

This should also fix the builds that are currently broken.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,183 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  typecheck:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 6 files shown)


daopunk (approved):
yay linters! def need all this stuff to handle more contributors

logicminds:
disabled for now as most of the fail, new pr will be required to fix all of the current issues.

PR #88, opened 2026-07-16
fix(security): harden session, migration, and export paths from review

logicminds:
## Summary

This PR closes the security-review-hardening gaps so the app can no longer be tricked into exposing secrets or mutating wallets after the session has locked.

**Before:** an idle-locked session still allowed EVM account enumeration and edits, migration could run out of memory or leave an account half-migrated after a crash, and clipboard exports could write a secret even if the session locked while the user was typing their PIN. The frontend also never re-checked the backend session when the app regained focus.

**After:** sensitive commands require an active session at entry and immediately before writing secrets, migration streams rows safely and can resume from an interruption, exports validate their inputs and clear the clipboard on failure, and the frontend re-verifies the session on focus and resume.

## Changes

- Added session-lock guards to every EVM account command with a heartbeat reset.
- Replaced the all-at-once migration scan with streaming table migration, a 10k-row safety gate, and an `in_progress` recovery marker.
- Fixed the export TOCTOU by snapshotting session state at entry and re-checking before the clipboard write, with the pending flag set before the write and cleared on failure.
- Switched key-derivation salt generation to `OsRng` and capped the export backoff shift to prevent overflow.
- Added `initSessionFocusChecks` so the webview debounces `checkSession` on `window.focus` and `document.visibilitychange`.
- Added tests for locked EVM commands, clipboard startup/shutdown cleanup, audit logging, backoff integration, malformed export requests, and focus/resume checks; configured Rust tests to run serially to avoid global-state races.

## Test plan

- `cd src-tauri && cargo test` → 233 passed
- `pnpm test` → 137 files, 1170 tests passed
- `pnpm check` → 0 errors

---


logicminds:
Closing in favor of three focused PRs that split the security-review-hardening work:\n\n- #95 — per-device salt and migration to v2\n- #96 — session manager, idle auto-lock, and migration gate\n- #97 — backend-mediated clipboard export for sensitive secrets\n\nThe unrelated docs and scratch notes were moved to the docs/out-of-band-planning branch.

PR #96, opened 2026-07-16
feat(session): add automatic idle lock and migration gate for sensitive operations

logicminds:
## Summary

This PR introduces automatic session idle lock and a migration gate so that sensitive account and treasury operations are only available while the user is actively using the app and after their account has migrated to the v2 per-device salt derivation.

Before this change, the encryption key stayed in memory indefinitely after unlock, and v1-derived accounts could still perform treasury/governance operations without migrating. After this change, the session locks after idle timeout, clears the encryption key, and requires re-authentication; legacy accounts must migrate before accessing wallet, squad, or governance features.

### Backend 
- Adds `session.rs` manager that tracks user activity via heartbeats, locks the session after idle timeout, and clears the in-memory encryption key.
- Adds a migration gate that blocks wallet, treasury, squad, and governance commands until the current account has migrated to the v2 key derivation.
- Integrates session heartbeat/lock into command handlers and the global state lifecycle.

### Frontend 
- Wires auth state to re-check session status on window focus and app resume, redirecting to the PIN/password unlock screen when locked.
- Adds idle lock controls and migration gate UI in `AppSettingsSection`.
- Adds tests for session lifecycle, idle timeout, heartbeat, and migration-gated paths.

## Test plan

- `cd src-tauri && cargo test` covers session lifecycle and migration gate.
- `pnpm test` covers frontend auth state, session heartbeat, and idle lock behavior.

---

Screenshot 2026-07-18 at 5 55 26 PM


src-tauri/src/evm/squad_sponsor_common.rs
```diff
@@ -27,6 +27,22 @@ fn require_parent_member_decision(
     }
 }
 
+/// Preflight: reject when this parent has no sponsor infra row in SQLite.
+pub fn require_sponsor_infra_for_parent<R: Runtime>(
```
daopunk: Pacto-Gov should be allowed to be deployed before the Sponsor, because when deployed together, the Pacto-Gov Hats tree is used as an init param for the Sponsor to base sponorship off of the Hats, rather than individual approved addrs

src-tauri/src/lib.rs
```diff
@@ -5590,6 +5560,12 @@ async fn add_mls_member_device(
     member_npub: String,
     device_id: String,
 ) -> Result<(), String> {
+    session::heartbeat();
+    if let Some(handle) = TAURI_APP.get() {
+        crate::migration::require_key_derivation_version_2_on_handle(handle)?;
+    } else {
+        return Err("App handle not initialized".to_string());
+    }
```
daopunk: can we make a helper function for this to not duplicate the if-else statements

daopunk (changes requested):

daopunk (approved):

Issue #102, opened 2026-07-17
chore: normalize Nostr squad-state gossip for dashboard parity


daopunk:
## Summary

Squad dashboard columns (EVM / hats / privileges / sponsored) can disagree across members when state arrives over Nostr/MLS rather than chain reads. Example: each member sees only their own shared roster EVM; peers stay **Not shared**.

**Immediate fix PR:** branch `fix/squad-evm-share-announcements-ingest` — `squad_member_evm_share` was routed to **announcements** but SQLite upsert only ran for the **inbox** virtual-bucket side-effect gate.

## Problem class

Chain-backed views stay consistent because RPCs are shared truth. Nostr/MLS automation messages need a **normalize-on-ingest** path into local SQLite + a **live store refresh** so every member’s dashboard converges.

## Scope for follow-up

Audit other MLS announces that mutate local tables but may be gated on the wrong virtual bucket (or lack frontend merge handlers), including:

- [ ] `squad_safe_updated` / Safe proposal automation vs ADR bucket rules
- [ ] `squad_contract_allowlist_updated`
- [ ] `squad_tracked_tokens_updated`
- [ ] Any other `apply_inbox_virtual_bucket_side_effects` inbox-only applies that producers send to `announcements`
- [ ] Backfill / `replayMlsAutomationSideEffects` coverage for historical messages
- [ ] Optional Rust emit (`squad_member_evm_updated`) instead of content-type parsing in the UI
- [ ] Integration test: ingest announcements-bucket `squad_member_evm_share` → `list_squad_member_evm` includes author

## Acceptance

- Two clients in the same squad: after A shares EVM, B’s Crew shows A’s address without restart (and vice versa).
- Same for any additional announce types fixed in the follow-up PR.
- Docs (`VIRTUAL_CHANNEL_ROUTING_ADR`, `DESIGN.md`) stay aligned with ingest gates.

## Related

- Design: `docs/communities/DESIGN.md` (roster EVM → #announcements)
- ADR: `docs/mls/VIRTUAL_CHANNEL_ROUTING_ADR.md` rule 6

daopunk:
closed by https://github.com/covenant-gov/pacto-app/pull/111 #111

PR #103, opened 2026-07-17
fix: persist peer roster EVM shares from announcements

daopunk:
## Summary

Before: when a member shared their squad EVM key, only their own Crew tab updated. Peers kept showing **Not shared**, even though the share appeared in #announcements chat.

After: peers persist `squad_member_evm_share` into SQLite on MLS ingest and refresh Crew maps live.

Root cause: shares are routed to the **announcements** virtual bucket, but the upsert only ran under the **inbox** side-effect gate.

## Changes

- Apply `try_apply_squad_member_evm_share` for announcements (and missing-bucket) ingest.
- On `mls_message_new` / message update, refresh roster EVM maps for parents whose #announcements group matches.
- Unit coverage for the subscription + sync helpers.

Follow-up for the broader Nostr/MLS dashboard gossip audit: #102

## Test plan

- [x] `pnpm exec vitest run src/lib/app/tauri-subscriptions.test.ts src/lib/dashboard/dashboard-data-sync.test.ts`
- [ ] Two clients: A shares EVM → B’s Crew shows A’s address without restart (and vice versa)
- [ ] Optional: open #announcements and confirm share card still appears

Made with [Cursor](https://cursor.com)

### DIFF
--- a/src-tauri/src/db.rs
+++ b/src-tauri/src/db.rs
@@ -2783,8 +2783,9 @@ pub async fn replay_automation_side_effects_for_chat<R: Runtime>(
 }
 
 /// Applies treasury (`squad_safe_updated`), governance (`governance_updated`), and roster EVM (`squad_member_evm_share`)
-/// side effects when the MLS chat row is classified into the **inbox** virtual bucket (routing ADR).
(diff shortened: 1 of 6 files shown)


daopunk:
closing in favor of #111

PR #125, opened 2026-07-21
feat: polish at-mentions ux in squad channels

logicminds:
## Summary

This PR rounds out the squad-channel @-mention experience so it behaves like a standard chat composer. Users can now mention members without first opening the members sidebar, autocomplete follows the cursor and flips above the input when room runs out, and the list stays tidy by hiding people already mentioned in the current message.

## What's new

- **Mentions work without the members panel** — the active channel's member list is loaded automatically, so the composer always has candidates.
- **Cursor-aware autocomplete positioning** — the popup opens next to the caret and flips above the line when the input is near the bottom of the viewport.
- **Trigger on any `@`** — autocomplete opens after any `@` character, not only at the start of the message or after whitespace.
- **Smart de-duplication** — already-mentioned members are hidden from the list; deleting a mention from the text restores the candidate.
- **Cleaner boundary matching** — `filterMentionsInText` no longer relies on `\b` word boundaries, which could match mid-word.

## Test plan

- `pnpm test` — 161 files, 1415 tests pass.
- `pnpm check` — 0 errors.
- `pnpm lint` — clean.

---

Screenshot 2026-07-21 at 1 18 13 PM
Screenshot 2026-07-21 at 1 18 20 PM
Screenshot 2026-07-21 at 1 18 55 PM


logicminds:
superseded by #129

Issue #141, opened 2026-07-23
A1: Add message reactions (DM + MLS) — backend already ready


logicminds:
## Summary
Vector exposes emoji reactions on every message; pacto has no UI for them. The Rust backend still implements reactions (NIP-25 kind=7) and returns them from \`get_message_views\`; the frontend simply does not render them or provide a way to add one.

## Backend commands already available
- \`react_to_message(reference_id, chat_id, emoji)\` — \`src-tauri/src/message.rs\`
- Reactions are already stored and returned by \`get_message_views\`.

## Frontend work required
1. Extend \`DmMessage\` type in \`src/stores/dm.ts\` to carry a \`reactions\` field.
1. Update \`normalizeDmPayload\` in \`src/lib/app/tauri-subscriptions.ts\` to preserve reactions from the backend payload.
1. Add a reaction chips row in \`src/components/dm/Message.svelte\` (grouped emoji + count, highlight the user's own reaction).
1. Reuse the existing emoji picker in \`src/components/dm/MessageInput.svelte\` as a reaction trigger.
1. Wire the picker selection to call the existing API wrapper for \`react_to_message\`.
1. Handle incoming reaction updates via the existing \`message_update\` event path.

## Acceptance criteria
- [ ] Users can add/remove a reaction on any plain DM or MLS channel message.
- [ ] Reactions render inline below the message bubble, grouped by emoji with count.
- [ ] The user's own reactions are visually distinct.
- [ ] Works for both DMs and MLS group channels.

## Effort / risk
**Effort:** S (UI-only, backend ready).  
**Risk:** Low — no protocol/crypto/backend changes.

logicminds:
fixed in #157

PR #157, opened 2026-07-24
feat: messaging enhancements — reactions, attachments, image viewer, sync status

### LINKED ISSUE #143: A3: Add attachment composer + rendering — backend file pipeline ready
## Summary
Vector supports drag-and-drop file/image/video/audio sending with a compression toggle and savings display. Pacto has no attachment UI at all, despite a complete file-pipeline backend inherited from the same fork.

## Backend commands already available
- \`file_message\`, \`file_message_compressed\`, \`send_file_bytes\`, \`send_cached_file\`, \`forward_attachment\` — \`src-tauri/src/message.rs\`
- \`download_attachment(npub, msg_id, attachment_id)\` — \`src-tauri/src/lib.rs\`
- Attachment metadata already flows through \`get_message_views\`; reply previews already show an "Attachment" placeholder.

## Frontend work required
1. Add a paperclip / drag-drop attachment target to \`src/components/dm/MessageInput.svelte\`.
1. Add a compression toggle that shows original vs. compressed size (backend enforces a 10% minimum-savings threshold).
1. Extend \`DmMessage\` / \`normalizeDmPayload\` to carry attachment metadata.
1. Render attachments in message bodies: inline images, file cards with name/size/type for other files.
1. Add open/save/download affordances for received attachments.

## Acceptance criteria
- [ ] Users can attach files via click, paste, or drag-and-drop in DMs and MLS channels.
- [ ] Image compression toggle shows original and resulting size.
- [ ] Sent images render inline; other files render as downloadable cards.
- [ ] Received attachments can be opened or saved.

## Effort / risk
**Effort:** M (UI + type plumbing).  
**Risk:** Low-to-medium — touches composer and message rendering but uses existing backend.

logicminds:
## Summary

This PR ships the messaging baseline: per-message reactions, file attachments with an in-app image viewer, reply previews, and per-conversation sync status indicators.

Before this change, the Rust backend already supported reactions, attachments, and sync state, but the Svelte frontend had no consistent way to add, display, or interact with them. Members had to paste URLs, restart the app to catch up, or could not react at all.

After this change, members can:

- React to any message from a long-press / right-click context menu and inline emoji picker.
- Send and preview image/file attachments from the composer.
- Tap images to open the in-app viewer with zoom, pan, and reveal-in-folder.
- See whether a DM or squad channel is still catching up in the thread header.

## Changes

- Added a shared per-message context menu (`MessageActionsMenu`) with React, Copy, and Reply actions.
- Added an inline reaction emoji picker in `Message` with search and recent emojis.
- Rendered aggregated reaction chips below the message bubble.
- Unified reply preview into `MessageInput` so both DMs and squad channels show the original author and content being replied to.
- Populated reply context (`replied_to_content`, `replied_to_npub`, `replied_to_has_attachment`) in the backend before sending, so both sides see the reference.
- Wired `sendFileBytes` for DMs and added attachment previews in the composer.
- Added byte-level MIME sniffing in the outgoing message path so Blossom uploads get the correct `Content-Type` instead of `application/octet-stream`.
- Added a sync status indicator to DM and squad thread headers.
- Added an in-app image viewer for attached images with zoom, pan, and reveal-in-folder.
- Added `MessageInput.test.ts` and updated `MessageActionsMenu.test.ts`.
- Added the implementation plan and requirements documents.

## Test plan

- `pnpm lint` — clean
- `pnpm check` — 0 errors (only pre-existing warnings)
- `pnpm test --run` — 170 files, 1485 tests passed
- `cd src-tauri && cargo test` — 312 tests passed

Close #143 #159 #141


logicminds:
Screenshot 2026-07-24 at 5 27 25 PM
Screenshot 2026-07-24 at 6 44 52 PM

daopunk (changes requested):
## Review: messaging v0.5 (#157)

**Verdict: request changes.** Solid feature surface and good helper/unit coverage, but text replies are not wired end-to-end, and two Tauri trust-boundary issues should be tightened before merge.

CI note (at review time): lint / typecheck / frontend-tests / e2e look green; treat backend-tests as still required.

---

### Ship-blocking

#### 1. Critical — text replies never attach `repliedTo`

Parents were updated to accept a reply id, but the composer never forwards it on text send.

- `MessageInput.svelte`: `onSend` / `onSendMentions` remain `(content) => …`; `handleSubmit` still calls `onSend(body)` / `onSendMentions(body, pruned)` with no `repliedTo`.
- Attachment path correctly uses `repliedTo ?? ''` in `sendPendingAttachment`.
- `+page.svelte` `handleDmSend(content)` still calls `sendDmMessage(id, content)` (API already supports `repliedTo`).
- `DmThread.svelte` / `ChatView.svelte` have `handleSend(…, repliedTo?)` ready, but never receive the id from `MessageInput`.

**Effect:** Reply chrome is cosmetic for text; only file replies are real. Contradicts the PR summary around reply previews / backend populate.

**Fix:** Thread `repliedTo` through `onSend` / `onSendMentions` → DM/channel handlers → `sendDmMessage` / group send. Add a unit test that reply + Enter forwards the id.

---

### High — security / privacy

#### 2. `save_attachment_as` trusts webview destination path

New command in `src-tauri/src/lib.rs`: source is constrained under `vector/`, but `dest_path` is caller-controlled; creates parents and `fs::copy`s plaintext there.

**Risk:** Compromised webview → arbitrary overwrite / exfil of decrypted attachments.

**Fix:** Open the save dialog in Rust (never accept a raw path from JS), or canonicalize + allowlist under Downloads/Documents; refuse `create_dir_all` outside that root.

#### 3. Asset protocol enabled over decrypted attachment dirs

`tauri.conf.json`: `assetProtocol.enable: true` and scope adds `$DOWNLOAD/vector/**`, `$DOCUMENT/vector/**` (CSP remains `null`).

**Risk:** Broadens local plaintext media readable via `asset://` if XSS exists in the webview. Intentional for the image viewer, but expands blast radius.

**Fix:** Keep scope minimal; prefer IPC/blob serving; tighten CSP over time.

#### 4. Link-preview fetch on message receive

`tauri-subscriptions.ts` `maybeFetchLinkPreview` fires on `message_new` / `message_update` for any `https://` content (in-memory dedupe only, no concurrency cap). Backend `fetch_msg_metadata` SSRF posture is pre-existing / unchanged.

**Risk:** Peer-forced egress + tracking; privacy-sensitive for this product.

**Fix:** Viewport/user-gated fetch; in-flight cap; harden backend (private IP / redirect / size limits) as follow-up.

---

### High / Medium — correctness

| Sev | Finding | Where |
|-----|---------|--------|
| High | Composer reply preview uses `msg.replied_to_has_attachment` (whether *that* message was a reply to an attachment), not whether **this** message has attachments | `DmThread.svelte`, `ChatView.svelte` `onReply` |
| High | Pending reaction never cleared on `reactToMessage` failure → emoji stuck until logout | `Message.svelte` + `reactions.ts`; parents `.catch` toast only |
| Medium | `LinkPreview` binds raw `og_image` / `favicon` as `` (text is escaped) | `LinkPreview.svelte` |
| Medium | MIME sniff only when extension → `octet-stream`; known wrong ext skips `infer` | `util.rs` `sniff_extension_and_mime` |
| Medium | `.svg` treated as image for send/preview | `attachment-composer.ts` / `attachment-display.ts` |
| Medium | Global `pendingFilePreview` + every mounted `MessageInput` registers the same drop handler | `attachment-composer.ts`, `attachment-drop.ts` |
| Medium | `ImageViewer` may leave `pointermove`/`pointerup` listeners if closed mid-drag | `ImageViewer.svelte` |
| Low | Remaining hardcoded English (toasts / some aria) despite i18n pass | Message / Input / ChatView / DmThread |

---

### What’s in good shape

- Attachment send path wires `repliedTo` end-to-end.
- Blossom ciphertext upload Content-Type → `application/octet-stream` is an improvement.
- Incoming filename sanitization + tests look careful.
- Reaction aggregation helpers and attachment display helpers have unit tests.
- i18n work cleared many raw-text warnings; MLS welcome copy localized.
- Frontend/e2e CI green; plan/requirements docs accompany the feature.
- `AppConfig` IPC is read-only; blossom dead-code removal is not a security regression.

---

### Recommended merge gate

Block merge until:

1. Text reply `repliedTo` wired + tested (**Critical**).
2. Composer preview uses `attachments?.length` (**High** UX).
3. Pending reactions cleared on failure (**High** UX).
4. `save_attachment_as` no longer accepts unconstrained JS paths (**High** security).

Treat asset-protocol scope + link-preview auto-fetch as merge-or-fast-follow depending on threat model; document the privacy tradeoff if shipping auto-fetch.

logicminds:
Thanks for the thorough review — addressed the merge-blocking items plus the rest of the actionable feedback in ce69bbd.

## Fixed

**1. Critical — text replies never attach `repliedTo`.** Confirmed: `MessageInput.svelte`'s `onSend`/`onSendMentions` signatures and `handleSubmit` weren't forwarding the component's own `repliedTo` prop, even though `DmThread.handleSend`/`ChatView.handleSendText`/`handleSendMentions` below were already wired to receive it. Threaded `repliedTo` through `onSend(body, repliedTo)`/`onSendMentions(body, pruned, repliedTo)`, and updated `handleDmSend` in `+page.svelte` to accept and forward it to `sendDmMessage`. Added a regression test in `MessageInput.test.ts`.

**2. `save_attachment_as` trusts webview destination path.** Moved the save dialog into Rust itself (mirroring the existing `select_custom_notification_sound` pattern in `audio.rs` — `tauri_plugin_dialog::DialogExt` + `blocking_save_file()`), so the command no longer accepts a `dest_path` string over IPC at all; the destination is chosen exclusively by a native OS dialog on the Rust side.

**3. Composer reply preview uses the wrong field.** `onReply()` in `ChatView.svelte`/`DmThread.svelte` checked `msg.replied_to_has_attachment` (whether `msg`'s *parent* had an attachment) instead of whether `msg` itself has attachments. Now checks `msg.attachments?.length`. The other `replied_to_has_attachment` usages (the quoted "replying to" header) were correct as-is and left untouched.

**4. Pending reaction never cleared on failure.** Both `ChatView.svelte`'s and `DmMessageRouter.svelte`'s `onReact` now call `clearPendingReactions(messageId)` in their catch blocks, so a failed reaction no longer permanently blocks retrying that emoji.

**5. `LinkPreview` binds raw `og_image`/`favicon` as ``.** Added an `isHttpUrl()` scheme allowlist — only `http:`/`https:` reach ``; anything else (`javascript:`, `file:`, `data:`, `asset:`) degrades to the existing no-image fallback. Added `LinkPreview.test.ts` (didn't exist before).

**6. `ImageViewer` may leave `pointermove`/`pointerup` listeners if closed mid-drag.** Added `onDestroy` cleanup removing both window listeners unconditionally, not just on a real `pointerup`. Added a regression test.

**7. Remaining hardcoded English (toasts / aria).** Wired every remaining literal `aria-label`/`title`/`showToast` string in `Message.svelte`, `MessageInput.svelte`, and `ChatView.svelte` to `$t(...)`, with new keys added to both `en/messaging.json` and `es/messaging.json`.

Validation: `pnpm vitest run` — 1700/1700 passing (189 files); `cargo test --lib` — 335/335 passing, 1 ignored (requires a live relay); `pnpm lint` clean.

## Not addressing

> MIME sniff only when extension → octet-stream; known wrong ext skips `infer`

`sniff_extension_and_mime`'s own doc comment (added in this same commit) states the intended behavior: "sniff the real extension and MIME type from file bytes **when the extension is missing or unknown**." The function does exactly that. Downstream, the sniffed value is a display hint for attachment rendering (image/video/document icon selection) — not an access-control gate: the network upload always declares `application/octet-stream` regardless (ciphertext), and rendering goes through ``/native players that don't execute embedded content. A mismatched declared extension produces a broken preview at worst, not code execution.

> `.svg` treated as image for send/preview

Both render paths (`MessageAttachment.svelte`'s poster tile and `ImageViewer.svelte`) use `` exclusively — browsers do not execute scripts embedded in an SVG loaded via `` (unlike `<object>`/`<iframe>`/direct navigation), so this doesn't create a script-execution vector in this codebase's rendering path.

> Global `pendingFilePreview` + every mounted `MessageInput` registers the same drop handler

Checked all three `<MessageInput>` mount points (`ChatView.svelte` ×2, `DmThread.svelte` ×1) — each pair is behind mutually-exclusive `{#if}`/`{:else}` branches, so only one `MessageInput` is ever mounted at a time in the reachable UI. `registerAttachmentDrop`'s reference-counted native listener is explicitly documented as a deliberate singleton, and `MessageInput`'s `onDestroy` already calls `clearPendingAttachment()` on unmount, so switching chats already clears the global store correctly.

## Needs a decision (open, not resolved)

Bundling these since your own "Recommended merge gate" already grouped them:

> Treat asset-protocol scope + link-preview auto-fetch as merge-or-fast-follow depending on threat model; document the privacy tradeoff if shipping auto-fetch.

Going to think through the threat-model tradeoff here rather than pick unilaterally — this affects the product's stated privacy posture, not just code correctness.

logicminds:
Follow-up on the open privacy/blast-radius decision from my last comment:

> Treat asset-protocol scope + link-preview auto-fetch as merge-or-fast-follow depending on threat model; document the privacy tradeoff if shipping auto-fetch.

Decision (0532066): keep the asset-protocol scope as-is (already limited to the app's own attachment directory, needed for the image viewer), but gate link-preview fetching on viewport visibility instead of firing eagerly on every message arrival/load — `requestLinkPreview` now only fires once a message's row actually intersects the viewport (`use:observeLinkPreview`, 200px lookahead), with fetches capped at 3 concurrent in-flight. This closes most of the privacy gap (off-screen/unread messages no longer leak read-behavior or IP to whoever controls an embedded link) and is also a straightforward lazy-load perf win — previews for messages the user never scrolls to are simply never fetched.

Same commit also renames the local attachment storage directory from the legacy `vector` codename to `pacto` (`$DOWNLOAD/pacto`, `$DOCUMENT/pacto`) in the asset-protocol scope and the Rust save/download paths, unrelated to your review but caught while touching this area.

logicminds:
Screenshot 2026-07-24 at 10 44 06 PM

Replies are not working as intended

daopunk (approved):

Issue #147, opened 2026-07-23
A7: Add per-relay health detail to Nostr settings


logicminds:
## Summary
Vector's relay list shows per-relay status, ping, and last-check detail in a popup. Pacto's \`NostrSettingsSection.svelte\` only supports add/remove; users can't see relay health.

## Backend state
- \`get_relay_metrics(url)\` and \`get_relay_logs(url)\` are registered Tauri commands in \`src-tauri/src/lib.rs\`.
- The comparison report notes we should first verify what the underlying \`nostr-sdk\` already tracks before adding new backend state.

## Frontend work required
1. Verify which relay stats are cheaply available from existing commands (connected/disconnected, last ping, error count, etc.).
1. Extend \`NostrSettingsSection.svelte\` (or add a small detail popover) to display per-relay health.
1. Add a refresh/last-checked affordance if the backend supports it.

## Acceptance criteria
- [ ] Each relay row in settings shows an expandable health detail.
- [ ] Detail includes connection status and any available latency/last-check info.
- [ ] No new backend polling unless necessary.

## Effort / risk
**Effort:** M (UI + verifying backend data).  
**Risk:** Low once data source is confirmed.

logicminds:
fixed in #169

PR #169, opened 2026-07-28
feat(relays): add per-relay health detail to Nostr settings

logicminds:
## Summary

Settings > Nostr previously only let users add/remove relays with no visibility into whether they're actually working. Each relay row now expands into a health detail panel: connection status, approximate ping/last-checked time, approximate data transferred, and a recent activity log — closing #147.

Before this change, a broken or misconfigured relay looked identical to a healthy one in the relay list. After this change, clicking a relay's disclosure toggle reveals live connection health, sourced from the existing `get_relay_metrics`/`get_relay_logs` commands with no new backend polling.

Since the PR was opened, review and live-app testing surfaced four more bugs — two of them pre-existing crashes/data-loss risks in the login/restore path, unrelated to the relay feature itself but caught while exercising it end-to-end — all now fixed on this branch.

## Changes

- **Backend:** two new instrumentation points feed the existing (previously unused) `RelayMetrics`/`RelayLog` state — `record_event_received` counts per-relay-per-delivery receipts (hooked off `RelayPoolNotification::Message` rather than the pool-deduplicated `::Event` stream, so every relay that delivers a copy of an event is counted, not just whichever relay was first), and `record_send_outcome` counts accepted/rejected sends at every existing `send_event`/`send_event_to` call site across `lib.rs`, `commons.rs`, and `mls.rs`.
- **Frontend API:** typed `getRelayMetrics`/`getRelayLogs`/`hasRelayHealthData` wrappers in `src/lib/api/relays.ts`.
- **UI:** an inline expandable detail panel per relay row in `NostrSettingsSection.svelte`, fetched on demand (not polled), with a manual refresh control that always stays available — including after a failed fetch, so a transient error never leaves the user stuck with no way to retry.
- **i18n:** new keys added to both `en` and `es` catalogs.

### Fixes since PR creation

- **Restore-from-recovery-phrase crash:** re-importing an account whose local profile DB already existed panicked in `crypto::internal_encrypt` ("Encryption key should be set"), killing the login command mid-flight and hanging the frontend on "Processing..." forever. `complete_login_from_keys` called `evm_accounts::ensure_ready` before `ENCRYPTION_KEY` is set in the recovery-phrase flow; now skipped when unset, since the frontend's own `encryptAndSaveEvmKey` call right after PIN entry already persists the same signing material.
- **Boot-time account scan could delete valid accounts:** `list_accounts()` runs on every app boot and removes any account directory whose DB has no `pkey` row (cleanup for abandoned mid-onboarding accounts). It collapsed *every* query failure — WAL lock contention with the main connection pool, transient I/O, mid-migration state — into the same "no pkey" signal as a genuinely empty account, so a busy database could get `remove_dir_all`'d. Query-result interpretation is now a pure, unit-tested classifier: only a definitive "no rows" is treated as deletable; any other error now skips the account instead of deleting it, and the raw connection gets a 2s busy timeout so ordinary lock contention resolves instead of erroring.
- **MLS welcome replay noise:** historical gift-wrap resync (runs every login) never marked processed MLS Welcome wrapper events as handled, so a permanently-failed welcome (e.g. sent to a pre-restore KeyPackage) re-attempted and re-logged the same non-actionable MDK error on every single login, forever. Wrapper events are now deduped through the existing discarded-giftwrap path once processed, whether they succeeded or permanently failed.
- **Relay detail panel overflow:** the new detail panel used `width: 100%` in an app with no global `box-sizing: border-box` reset, so content-box sizing added padding/border on top of the full container width and overflowed past the relay row above it. Switched to the block default (`width: auto`), which self-adjusts correctly.
- **Refresh button gave no feedback:** `get_relays`/`get_relay_metrics`/`get_relay_logs` are fast local invokes that routinely resolved in under one animation frame, so the refresh icon's spin was too brief to register — clicking refresh looked like it did nothing. Both the relay-list and per-relay-detail refresh now hold their spin for a 400ms floor regardless of how fast the backend responds. Also dropped a redundant, mislabeled "Loading relays…" text node next to the per-relay refresh icon that flashed in and out in that same sub-frame window; the icon-only spin now carries the loading signal per this repo's icon-only-refresh convention.

## Test plan

- `cargo test --lib` — 351 passed, 0 failed, 1 ignored (includes 7 new `relay_metrics_tests` and a new unit test for the account-scan pkey-query classifier).
- `pnpm vitest run src/lib/api/relays.test.ts` — 18/18 passed.
- `pnpm check` / `pnpm lint` — clean.
- Manual end-to-end verification via the Tauri MCP bridge against the running dev app:
  - Relay health panel: expanded a relay row and confirmed real connection status, approximate bytes/events, and recent-activity log render from live backend data; confirmed independent per-row expand/collapse state and the refresh control.
  - Restore-from-recovery-phrase panic: reproduced the exact crash against an existing account directory, applied the fix, restarted the dev app, and confirmed the same restore flow completes cleanly through to the Commons feed.
  - Refresh spinner: created a fresh account, opened Settings > Nostr, and instrumented the live DOM to poll the icon's spin state on click — confirmed the list-level and per-relay refresh both hold `is-spinning` for a guaranteed ≥400ms window before clearing, and confirmed no stray loading text renders in the per-relay detail head at any point during the fetch.
- Full internal code review pass (correctness, testing, reliability, maintainability, api-contract, adversarial personas) with fixes applied for a receive-count undercount and a refresh-failure retry dead end; two lower-severity architectural findings filed as follow-up issues (pacto-app-1e6, pacto-app-4ps) rather than expanding this PR's scope.

---

## demo
Screenshot 2026-07-28 at 11 20 25 AM


src-tauri/src/lib.rs
```diff
@@ -1429,6 +1438,17 @@ async fn handle_event(event: Event, is_new: bool) -> bool {
                     .await
                     .unwrap_or(false);
 
+                    // Mark this wrapper event as handled so historical resyncs (every login)
+                    // don't re-unwrap and re-attempt it forever, whether it succeeded or
+                    // permanently failed (e.g. a Welcome sent to a pre-restore KeyPackage).
+                    if let Some(handle) = TAURI_APP.get() {
+                        let _ = db::record_discarded_giftwrap(handle, &wrapper_event_id).await;
```
daopunk: **High — MLS welcome permanently discarded on any failure**

**Problem:** After every welcome attempt, this always writes the wrapper id to `discarded_giftwraps` (and the in-memory cache), whether `processed` is true or false. `wrapper_event_exists` treats discarded ids as already handled, so historical resync (`is_new == false`, every login) never retries.

That is too broad: `false` also covers transient cases that never got a permanent MDK verdict — `MlsService::new_persistent` / `engine()` failing, `TAURI_APP` missing in the blocking task, `spawn_blocking` join failure, or an unknown `process_welcome` error. A first-attempt init blip then permanently loses the invite on later logins.

Permanent MDK replay failures (`"missing welcome for processed welcome"` / `"processed welcome not found"`) *should* be marked handled; success should too. Everything else should stay retryable.

**Recommended fix:** Only call `record_discarded_giftwrap` / cache insert when:
1. `processed == true`, or
2. the error is one of the two known permanent MDK strings.

Do **not** discard when MLS service/engine failed to open or the error is unknown/transient. Prefer a small outcome classifier (`Processed` / `PermanentFailure` / `TransientFailure`) from the blocking task, with a unit test covering those three cases.

src-tauri/src/account_manager.rs
```diff
@@ -417,3 +461,35 @@ mod is_in_flight_account_tests {
         assert!(!is_in_flight_account(orphan, None, None));
     }
 }
+
+#[cfg(test)]
+mod classify_pkey_query_result_tests {
```
daopunk, line 498: **Medium — missing regression test for the current-account inclusion bug**

**Problem:** `classify_pkey_query_result` and `is_in_flight_account` are well tested, but nothing asserts the policy that caused the Create Account screen regression: *current/pending with a valid pkey must still be included in the returned list* (cleanup-only protection, not list exclusion). That exact bug shipped once via an early `continue` on in-flight accounts.

**Recommended fix:** Extract a pure helper, e.g. `(has_pkey, in_flight) → Include | Delete | Skip`, and unit-test:
- `has_pkey=true` → Include (even when in-flight)
- `has_pkey=false, in_flight=false` → Delete
- `has_pkey=false, in_flight=true` → Skip
- query `Err` → Skip

Keeps coverage off the global account statics / real app-data dir.

src/components/settings/NostrSettingsSection.svelte
```diff
@@ -233,43 +304,114 @@
...
-                {#if relay.is_default}
-                  <span class="nostr-relay-badge">{$t('settings.defaultRelayBadge')}</span>
-                {/if}
+          <li class="nostr-relay-row-wrap">
+            <div class="nostr-relay-row">
+              <button
+                type="button"
+                class="nostr-relay-detail-toggle"
+                aria-expanded={openUrls.has(relay.url)}
+                aria-controls="nostr-relay-detail-{relay.url}"
```
daopunk: **Low — invalid HTML `id` / `aria-controls` from raw relay URLs**

**Problem:** `id="nostr-relay-detail-{relay.url}"` and the matching `aria-controls` embed values like `wss://relay.damus.io`, which contain `:` / `/` — invalid/fragile as HTML IDs and weak for accessibility wiring.

**Recommended fix:** Use a stable sanitized slug (e.g. hash of the URL, or host+path with non-alphanumerics replaced) for both `id` and `aria-controls`, keeping the real URL only in visible copy / invoke args.

daopunk (changes requested):
Thanks for the relay health work and the important login/account fixes caught along the way — those look solid (especially the pkey classifier and the current-account inclusion fix).

Requesting changes for **#1** (MLS welcome discard scope). **#2** and **#3** are smaller follow-ups called out inline.

daopunk (approved):

src-tauri/src/lib.rs
logicminds: > That is too broad: `false` also covers transient cases that never got a permanent MDK verdict — `MlsService::new_persistent` / `engine()` failing, `TAURI_APP` missing in the blocking task, `spawn_blocking` join failure, or an unknown `process_welcome` error. A first-attempt init blip then permanently loses the invite on later logins.

Addressed in c8bb0e2. You read it exactly right — all five of those paths collapsed into the same bare `false`, and the discard was unconditional below it.

The `spawn_blocking` closure now returns a `WelcomeOutcome` (`Processed` / `PermanentFailure` / `TransientFailure`) instead of a bool, and the permanent-vs-transient decision lives in one pure function, `classify_welcome_error(msg: &str)`, which treats only the two known MDK replay strings (`"missing welcome for processed welcome"`, `"processed welcome not found"`) as permanent. MLS service/engine init failure, a missing `TAURI_APP`, a join failure, and any unrecognized `process_welcome` error all classify as `TransientFailure`.

`db::record_discarded_giftwrap` and the `WRAPPER_ID_CACHE` insert are now gated on `outcome.should_mark_handled()`, true only for `Processed` and `PermanentFailure` — so a transient failure leaves the wrapper id unrecorded and the welcome gets retried on the next historical resync. Success and the two permanent strings still mark it handled exactly once, still without log spam, and the `mls_invite_received` emit plus `return true` still fire only on `Processed`.

Added `welcome_outcome_tests` covering the three-way classification and the `should_mark_handled()` truth table.

One deliberate exception: the separate `unsigned_opt == None` branch ("Failed to convert rumor to UnsignedEvent") still discards unconditionally. That's a deterministic parse failure against a fixed event payload, so retrying can only ever fail the same way — permanent by construction.

PR #203, opened 2026-08-07
fix: update gate freeze race

daopunk:
## Summary

Closes a race in the mandatory update gate where a late remote `minimum_compatible_version` verdict could flip the app to the non-dismissible block screen after auth had already been allowed to proceed — leaving an authenticated session stuck behind the gate until relaunch.

**Before:** `awaitGateBeforeAuth` could return `clear` (including on the fail-open timeout) while the remote check was still pending. Until `freezeGate()` ran after unlock/create/import completed, a late remote block could still apply mid-auth and then freeze as blocked over an authenticated session.

**After:** The gate commits (freezes) at the `awaitGateBeforeAuth` boundary before returning clear or blocked, so a late remote verdict cannot interrupt mid-auth. Success-path `freezeGate()` stays as an idempotent belt-and-suspenders call, moved to immediately after auth flags are set with no intervening awaits.

## Changes

- Freeze inside `awaitGateBeforeAuth` before returning, so R9 holds for the whole auth attempt (not only after success).
- Reorder `freezeGate()` in create/import/unlock to run right after `isAuthenticated` / `currentUser`, before any post-auth await.
- Add a regression test: timeout clear + late blocking remote leaves the gate clear.

## Test plan

- [x] `pnpm exec vitest run src/lib/updater/update-gate.test.ts src/stores/auth.test.ts` (61/61)

---

### DIFF
--- a/src/lib/updater/update-gate.test.ts
+++ b/src/lib/updater/update-gate.test.ts
@@ -261,6 +261,31 @@ describe('awaitGateBeforeAuth', () => {
       vi.useRealTimers();
     }
   });
+
+  it('freezes on clear so a late remote cannot flip to blocked mid-auth', async () => {
+    vi.useFakeTimers();
+    try {
+      const { promise: remotePromise, resolve: releaseRemote } =
+        Promise.withResolvers<FakeUpdate | null>();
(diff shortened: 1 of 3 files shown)


logicminds (approved):
LGTM

PR #231, opened 2026-08-09
feat: expandable chat input

### LINKED ISSUE #212: Composer: auto-expand chat input so multiline drafts are fully visible
## Problem

The message composer clips multiline drafts. Users cannot see the full note before sending — top/bottom lines are cut off inside the input.

Reported against the DM/channel composer (`MessageInput`): long wrapped text sits in a short field with vertical clipping.

## Current behavior

`src/components/dm/MessageInput.svelte`:

- `<textarea rows="1">`
- CSS: `max-height: 120px`, `resize: none`, `overflow-y: auto`
- No auto-grow on input (`scrollHeight` → height)

So the field stays ~single-line tall (plus padding) until/unless scrolling kicks in inside a short box — easy to miss content and hard to proofread.

## Expected behavior

As the user types or pastes multiple lines / wrapped text:

1. The composer **grows with content** so the full draft is visible (up to a sensible max).
2. Beyond the max, it scrolls *inside* an already-expanded field (not a tiny clipped strip).
3. Enter still inserts a newline (existing behavior); Send remains explicit / existing shortcut rules.
4. Shrinking back when text is deleted.

Match common chat UX (Slack/iMessage/Telegram-style growing composer).

## Acceptance criteria

- [ ] Multiline / wrapped drafts are fully readable before send (no clipping of first/last lines at normal lengths)
- [ ] Height auto-grows with content up to an agreed max (raise or keep ~120px only if product wants a low cap — prefer taller, e.g. ~30–40% of chat pane or ~200–240px+, then scroll)
- [ ] Works in DM and channel composers that share `MessageInput`
- [ ] Does not break mention picker placement, emoji panel, or attachment strip layout
- [ ] Keyboard / a11y: focus and caret remain correct after resize

## Notes

- Fix likely: on `input` (and mount), set `textarea.style.height = 'auto'` then `height = min(scrollHeight, max) + 'px'`.
- Consider shared utility if other composers exist.

daopunk:
## Summary

Multiline drafts in the DM/channel composer were clipped in a short field, so you couldn’t see the full note before sending.

The shared `MessageInput` now grows with content up to 240px (then scrolls), shrinks when text is cleared, and stays single-line inline when empty.

Closes #212.

## Test plan

- [ ] `pnpm exec vitest run src/components/dm/MessageInput.test.ts`
- [ ] Type/paste several lines in a DM and a channel — field expands; past max scrolls; delete/send shrinks back to one line
- [ ] Mention picker and emoji panel still open above the composer

---


daopunk:
Screenshot 2026-08-08 at 8 47 55 PM
Screenshot 2026-08-08 at 8 47 41 PM
Screenshot 2026-08-08 at 8 47 28 PM

PR #234, opened 2026-08-09
fix(mls): clean up chat rows and cursor on voluntary group leave

### LINKED ISSUE #219: Voluntary MLS group leave doesn't clean up chat/message DB rows
# Problem Description
`MlsService::leave_group()` (`src-tauri/src/mls.rs`) purges the local `mls_groups` metadata row for a channel the user voluntarily leaves, but does **not** delete that channel's `chats` row or its message history from the account SQLite DB. This is the existing behavior on the success path, not something introduced by #218 (the fix there only made a previously-failing "state lost" leave take the same success path).

Contrast with the involuntary eviction path, `cleanup_evicted_group()`, which explicitly:
1. Removes the chat from in-memory `STATE`
2. Deletes the chat row from the DB (`db::delete_chat`)
3. Drops the MLS event cursor for the group

A voluntary leave does none of the above beyond metadata removal — no `STATE.chats` retain, no `db::delete_chat`, no cursor cleanup.

# Expected Behavior (to be decided)
Not obviously a bug — could be intentional (user keeps local chat history as a personal archive after leaving) or an oversight (orphaned chat/message rows and cursor accumulate for a group the app otherwise treats as gone; `list_mls_groups` semantics around evicted-vs-left groups may also assume symmetric cleanup). Raising to get a product/eng decision on which behavior is correct, and if cleanup is wanted, whether it should mirror `cleanup_evicted_group()` exactly (including cursor removal) or be scoped differently (e.g. keep messages, drop only the cursor).

# Technical Context
- **Component:** Nostr MLS integration / Squad management, `src-tauri/src/mls.rs`.
- **Relevant code:** `MlsService::leave_group()` local-metadata-only cleanup vs. `MlsService::cleanup_evicted_group()`'s full cleanup (both in `src-tauri/src/mls.rs`).
- **Found while investigating:** #216 / #218.

logicminds:
Leaving an MLS group only removed the local `mls_groups` metadata row; the channel's chat, its messages, and the group's Nostr event cursor stayed behind in the database and in-memory `STATE`, unlike the involuntary-eviction path (`cleanup_evicted_group`) which already handles all three. `MlsService::leave_group()` now mirrors that cleanup: drops the chat from `STATE.chats`, deletes the chat/message rows via `db::delete_chat`, and removes the stale `mls_event_cursors` entry (keyed by both the wire and engine group ids, since callers pass either).

Fixes #219

Validation: `cargo build --lib` and `cargo test --lib mls::` pass (13/13, 1 ignored requiring a live relay); `cargo fmt -- --check` shows no diff for the changed file.

### DIFF
--- a/docs/mls/EVICTION_AND_LEAVE.md
+++ b/docs/mls/EVICTION_AND_LEAVE.md
@@ -30,7 +30,8 @@ When **this client** is removed, sync or live **444** handling detects eviction-
 
 ## 3. Voluntary leave (`leave_mls_group`)
 
-- Engine emits a **leave proposal**; Pacto publishes it, then **removes group from local metadata** for the leaver and emits **`mls_group_left`** — the group **disappears from the leaver’s list immediately**.
+- Engine emits a **leave proposal**; Pacto publishes it, then **removes the group row from `mls_groups`** (not merely marks it evicted) for the leaver and emits **`mls_group_left`** — the group **disappears from the leaver's list immediately** and stays gone even after a later sync.
+- Mirrors **`cleanup_evicted_group`**'s local cleanup: drops the chat from **`STATE`**, **deletes the chat/messages from DB**, and clears the **MLS event cursor** (keyed by both wire and engine group id).
 - **The group continues** for remaining members; MLS tree updates without the leaver.
 
 **Admin handoff:** Pacto does **not** currently expose “add admin” / “transfer MLS admin”. Only the creator is admin at creation. If the **creator leaves**, the MLS group may end up with **no admins** → **no further kicks** from MLS until MDK + app support admin updates. Squad-level roles (e.g. Hats) are a separate product layer.
--- a/src-tauri/src/db.rs
+++ b/src-tauri/src/db.rs
@@ -3726,7 +3726,10 @@ pub async fn save_chat<R: Runtime>(handle: AppHandle<R>, chat: &Chat) -> Result<
 }
 
 /// Delete a chat and all its messages from the database.
-/// `chat_identifier` is the npub for DMs or group_id for MLS; events are CASCADE deleted.
+/// `chat_identifier` is the npub for DMs or group_id for MLS. Deletes the `chats`
+/// row and its catch-up entries; connections never enable `PRAGMA foreign_keys`,
+/// so the schema's `ON DELETE CASCADE` on `events`/`messages` never actually
+/// fires — cleaning up those orphaned rows is tracked separately.
 pub async fn delete_chat<R: Runtime>(
     handle: AppHandle<R>,
     chat_identifier: &str,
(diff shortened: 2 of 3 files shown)


daopunk (changes requested):
## Audit report

Voluntary leave now drops in-memory chat, calls `db::delete_chat`, and clears `mls_event_cursors` (wire + engine ids). That part matches the #219 symptom and is stricter than `cleanup_evicted_group` on cursor keying.

The leave path still does **not** durably remove MLS membership metadata, so the new chat/cursor cleanup can be undone by sync. Commit wording also overstates what happens today.

---

### Blocking: `mls_groups` row is not actually removed

`leave_group` does `retain` + `write_groups`, but `db::save_mls_groups` is **upsert-only** (`INSERT OR REPLACE`). There is no `DELETE FROM mls_groups` in the backend.

So after leave:
- the `mls_groups` row remains with `evicted = false`
- multi-group sync still selects `!evicted` groups (`lib.rs`)
- sync can recreate chat state after this PR’s `delete_chat`

Eviction works because it sets `evicted = true`. Leave’s “remove metadata” claim is inaccurate on main and on this branch.

**Suggested fix (prefer 1):**
1. Add a real delete (e.g. `db::delete_mls_group` / delete-missing-ids in `write_groups`) so leave removes the row; **or**
2. Mark `evicted = true` on voluntary leave if product wants soft-leave / re-invite parity with eviction.

Until one of those lands, please don’t treat #219 as fully fixed.

---

### Shared purge / cascade wording (related: #233)

Commit text says `delete_chat` “cascades its messages.” Account connections never enable `PRAGMA foreign_keys`, so SQLite CASCADE does not run. `delete_chat` currently deletes the `chats` row (and catch-up entries) and can leave `events` orphans — same gap for eviction.

**Suggested fix:**
- Drop the cascade claim from the commit/PR description.
- Keep calling shared `db::delete_chat` (do **not** invent a DM+MLS deletion module).
- Prefer landing / rebasing after [#233](https://github.com/covenant-gov/pacto-app/pull/233), which hardens `delete_chat` with an explicit `DELETE FROM events` (+ legacy `messages`). That shared seam is enough; DM cutoffs stay DM-only, MLS metadata/cursors stay in `mls.rs`.

---

### Non-blocking follow-ups (same PR preferred)

1. Update [`docs/mls/EVICTION_AND_LEAVE.md`](docs/mls/EVICTION_AND_LEAVE.md) §3 once leave’s durable metadata + chat/cursor cleanup is correct.
2. Add a focused regression test: after leave, no chat row, no cursor for wire/engine ids, and group is absent **or** `evicted` so sync skips it.
3. Optional only: tiny helper for `STATE.retain` + `delete_chat` after #233/#234 both land — not required for merge.

---

### Ask

Please fix durable `mls_groups` removal (or `evicted=true`), correct cascade/metadata wording, and preferably rebase onto #233’s `delete_chat` hardening so leave/eviction actually purge events.

logicminds:
> `leave_group` does `retain` + `write_groups`, but `db::save_mls_groups` is **upsert-only** (`INSERT OR REPLACE`). There is no `DELETE FROM mls_groups` in the backend.

Fixed in ecd5a3d — went with option 1 (real delete). `write_groups` callers always pass the complete current group list, so `save_mls_groups` now diffs against what's persisted and deletes any row missing from the new list before upserting, all in one transaction. Confirmed the three call sites (`create_group`, the metadata-timestamp update in `add_member`, and `leave_group`) all already pass the full list, so this is a safe semantic tightening, not a behavior change for them. Added `replace_mls_groups_conn_deletes_group_missing_from_new_list` / `_upserts_kept_group_fields` unit tests against an in-memory migrated DB.

> Commit text says `delete_chat` "cascades its messages." Account connections never enable `PRAGMA foreign_keys`, so SQLite CASCADE does not run.

Confirmed — grepped every connection setup path and `PRAGMA foreign_keys` is never set. Fixed the doc comment on `delete_chat` to state that plainly instead of claiming cascade behavior that doesn't happen. Left the actual `events` orphan cleanup out of this PR per your suggested sequencing — that's #233's `DELETE FROM events` hardening, and duplicating it here risks the two PRs fighting over the same code. Will rebase onto #233 once it lands.

> Update `docs/mls/EVICTION_AND_LEAVE.md` §3 once leave's durable metadata + chat/cursor cleanup is correct.

Fixed — §3 now describes the row actually being deleted from `mls_groups` (not just evicted) and the chat/cursor cleanup this PR already added.

> Add a focused regression test: after leave, no chat row, no cursor for wire/engine ids, and group is absent or evicted so sync skips it.

Fixed for the group-removal half (the part this PR's bug was in) via the `replace_mls_groups_conn` tests above. The chat-row/cursor-removal half of that assertion needs the full engine+relay harness (`kim_group_with_saul`) that the other `leave_group` tests in `mls.rs` use — didn't duplicate that here since the DB-level fix is the piece that was actually broken; happy to add the full end-to-end test as a follow-up if you'd rather have it in this PR.

> Optional only: tiny helper for `STATE.retain` + `delete_chat` after #233/#234 both land — not required for merge.

Agreed this is optional and blocked on #233 landing first, so leaving it for then rather than guessing at the shared shape now.

`cargo test --lib` passes (595 passed, 2 pre-existing relay-dependent tests ignored).

daopunk (approved):

Issue #237, opened 2026-08-09
tech-debt: worktree-local dev sandboxes (isolated ports + seeded data) for parallel development


logicminds:
## Problem

`pnpm tauri dev` / `make dev` has two structural single-instance assumptions that don't hold once multiple git worktrees are being developed against at the same time:

1. **Fixed, `strictPort` dev server ports.** `vite.config.ts:56` binds the Vite dev server to `1420` with `strictPort: true` (no fallback), and the HMR websocket is pinned to `1421` (`vite.config.ts:63`). Two `make dev` processes running from two different worktrees at once hard-fail on port bind — genuine parallel dev sessions aren't possible today, full stop.
2. **Empty, per-branch-keyed sandbox data.** `dev-sandbox` (throwaway) and `dev-account`/`dev-buddy` (fixed, shared identities) are the only persistent options, and none of them are worktree-aware or pre-populated. A quick stopgap in this workspace changed the plain `dev` target to key `PACTO_TEST_SANDBOX_ROOT` off the current git branch (falling back to the real OS account dir only on `main`) so branch-hopping in a single worktree can't trip the storage-format update gate (`src-tauri/src/storage_format.rs`, see `docs/build/OPERATOR_UPDATES.md`'s "Recovering a machine the local storage-format check blocked" section for the failure mode). That's an improvement but not a real fix: every new worktree/branch starts from a **completely empty account** — no squads, no DMs, no wallet state — so there's nothing to look at until someone manually creates an account and repopulates it by hand, every single time.

## Why this gets worse with worktrees

Worktrees mean several branches under active, simultaneous development on one machine. That multiplies both failure modes: more concurrent `make dev` sessions wanting the same two ports, and more fresh sandboxes that start blank and need manual re-seeding before they're useful for "does this actually work" checks.

## What we need

A worktree-local dev sandbox that is both **isolated** and **useful out of the box**:

- **Per-worktree port allocation.** Derive a free/deterministic port pair for the Vite dev server + HMR from the worktree (e.g. hash of `git rev-parse --show-toplevel`, or scan for a free port and print it), and thread it through `vite.config.ts` / `TAURI_DEV_HOST` / `src-tauri/tauri.conf.json`'s `devUrl` so two worktrees can run `make dev` at the same time without touching each other.
[...]

PR #239, opened 2026-08-09
chore: isolate make dev's data dir per git branch

logicminds:
## Summary

`make dev` used a single shared local data directory regardless of git branch. Switching branches with different local database migrations tripped the storage-format update gate: a newer branch's schema, once written to that shared account, blocked an older branch's build from opening the same data — every account on the machine, not just one.

`make dev` now derives its `PACTO_TEST_SANDBOX_ROOT` from the current git branch. Data stays persistent (you still get a real account to look at, unlike the throwaway `dev-sandbox` target) but is isolated per branch, so branch-hopping can never trip the gate again. `main` is a deliberate exception — it keeps using the real OS app-data directory, so the existing dev account and PIN nobody wants to lose keep working exactly as before.

## Changes

- `Makefile`: `dev` target branches on `git rev-parse --abbrev-ref HEAD`; non-`main` branches get `test_fixtures/dev-branch-<slug>` (persistent, not swept by `make clean`).
- `docs/build/OPERATOR_UPDATES.md`: added a prevention note to the existing storage-format-gate recovery section pointing at this behavior.

## Test plan

- `make -n dev` dry-run confirms correct expansion for both the `main` and non-`main` branches.
- Verified branch-name sanitization (slashes, spaces) produces a clean filesystem-safe slug with no stray trailing characters.

## Follow-up

This is a stopgap, not the full fix. It doesn't solve running two `make dev` instances concurrently across worktrees (fixed `strictPort` Vite ports at 1420/1421 still block that), and a fresh worktree/branch still starts from an empty account. Tracked in #237.

---

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,7 +5,7 @@ help:
 	@echo "Pacto — available make targets"
 	@echo ""
 	@echo "  install      install Node/Rust dependencies"
-	@echo "  dev          run the desktop app in development mode"
+	@echo "  dev          run the desktop app in development mode (auto-isolated per branch, except main)"
 	@echo "  dev-sandbox  run the desktop app against a throwaway account for MCP-driven UI verification"
 	@echo "  dev-account  run the desktop app against the persistent, reusable primary test account"
 	@echo "  dev-buddy    run the desktop app against the persistent, reusable secondary test account"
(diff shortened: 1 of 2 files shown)


logicminds:
superceded by #244

Issue #173, opened 2026-07-28
Add a "Route Traffic Through Tor" setting using an embedded Arti client


logicminds:
## Summary

Add a new user-facing preference, **Route Traffic Through Tor**, that routes all outbound TCP traffic (Nostr relay connections, Blossom uploads, link previews, image fetches) through the Tor network via an embedded Arti client.

**Description (as it should read in Settings):**

> **Route Traffic Through Tor**
> When enabled, Pacto routes all TCP traffic (Nostr relays, Blossom uploads, link previews, image fetches) through the Tor network using an embedded Arti client. This hides your IP address from relays and remote servers, at the cost of slower connections (Tor circuits add latency).
>
> Tor and the Tor logo are trademarks of The Tor Project; all rights reserved. More information at torproject.org. Pacto is not endorsed or sponsored by, or affiliated with, The Tor Project.

## Current behavior

There is no Tor/SOCKS/proxy support anywhere in the codebase today (`src-tauri/Cargo.toml` has no `arti-client`/`arti` dependency). Every outbound connection is made directly, via several independently constructed clients:

- **Nostr relays (WebSocket):** `nostr_sdk::Client::builder()...build()` in `src-tauri/src/lib.rs` (two sites: initial login, `complete_login_from_keys`), connecting directly to `TRUSTED_RELAYS` and any user-added relays.
- **Blossom uploads/downloads:** `reqwest::Client::builder()` in `src-tauri/src/blossom.rs` (`upload_blob_with_progress_and_failover`), used for attachment and profile-media uploads to `get_blossom_blob_servers()` / `get_blossom_media_servers()`.
- **Link previews:** `reqwest::Client::builder()` (and a second `Client::builder()`) in `src-tauri/src/net.rs`, used by `fetch_site_metadata` (see #172) and a relay-health-check helper.
- **Image fetching:** `reqwest::Client` in `src-tauri/src/image_cache.rs` (static `HTTP_CLIENT` via `once_cell::Lazy`).
- **Voice transcription download:** `reqwest::Client::builder()` in `src-tauri/src/whisper.rs`.
- **EVM RPC calls:** `reqwest::Client::builder()` in `src-tauri/src/evm/sponsor_userop.rs` (bundler JSON-RPC) and `src-tauri/src/evm/wallet_prices.rs` (price oracle `eth_call`), plus whatever `alloy` providers construct internally for on-chain reads/writes.

[...]

PR #243, opened 2026-08-10
feat: add sticker packs and an opt-in klipy gif picker to composer

### LINKED ISSUE #200: feature: GIF/meme picker in message composer (Telegram-style)
## Summary

Add a GIF/meme picker to the DM and squad channel composer, matching the Telegram pattern in the attached screenshot: a search box, a tabbed panel (`Emoji` / `Stickers` / `GIFs`), and a scrollable grid of animated results that insert directly into the message on tap — no manual file picking.

## Current state

`src/components/dm/MessageInput.svelte` already has the two pieces this feature would sit beside:

- An **emoji panel** (`emojiPanelOpen`, `.emoji-panel`, lines ~769–908) triggered by a smiley-face button next to the textarea, with search (`emojiSearchQuery`) and a recent/browse grid. There is no `Stickers` or `GIFs` tab today — only emoji.
- An **attachment menu** (`attachmentMenuOpen`, `.attachment-menu`, lines ~717–767) for `Photo/Video`, `File`, and `Take Photo`, which funnels into `PendingFileAttachment` → `send_file_bytes` (see `docs/messaging/ATTACHMENTS.md` §6).

A GIF result is conceptually a third source alongside "type a message" and "attach a file" — it needs its own path, not a shoehorn into either existing one.

## Proposed behavior

1. Add a `GIFs` tab to the existing emoji/sticker panel (or a dedicated picker opened from the same toolbar), with a search input (debounced) and an infinite-scroll/paginated grid of thumbnails, matching the reference screenshot's layout.
2. Selecting a result sends it as the message immediately (Telegram behavior) rather than inserting a token into the text field — a GIF is a message, not composer text.
3. Support a "trending/recent" grid when the search box is empty, and a `no results` state.
4. Respect `disabled`/`isSendingAttachment` composer states the same way the attachment and emoji buttons already do.
5. Fully translated: new strings go in `src/lib/i18n/locales/{en,es}/messaging.json` under the `messaging.messageInput.*` namespace, consistent with `emojiTab`, `emojiPicker`, etc.

## Open design questions to resolve during implementation

- **Source of GIFs.** Pacto has no server backend (per `AGENTS.md`/`README.md` — Nostr + EVM only). A Telegram-style catalog needs a third-party API (Giphy, Tenor, or similar), which is a new network dependency and a new trust boundary:
  - Search queries would leave the device to a GIF provider, associated with the user's IP (and, if using an API key tied to app identity, the app instance). This needs a privacy note in `docs/`, similar to the Blossom disclosure in `docs/messaging/ATTACHMENTS.md` §4.
  - Requires a new operator secret (e.g. `GIPHY_API_KEY`/`TENOR_API_KEY`), which per this repo's conventions is a Rust-side `std::env` var, *not* a `VITE_*`/`ALCHEMY_*` client-exposed one — the search call should be proxied through a new `#[tauri::command]` so the API key never ships in the webview bundle.
- **Delivery mechanism.** Once a GIF is picked, how does it reach the peer?
  - **Option A — remote URL rumor:** send a kind-15-style rumor containing just the provider's CDN URL. Simplest, but leaks the GIF choice (and provider CDN correlation) to anyone who can see the rumor content pattern, and breaks if the provider deletes/rate-limits the URL — same "third-party retention" caveat already logged in `docs/messaging/ATTACHMENTS.md` §7 for Blossom, but now for a second uncontrolled host.
  - **Option B — fetch, then reuse the existing attachment pipeline:** download the GIF bytes client-side and pass them through the same `send_file_bytes` flow used for file attachments (sniff → hash → AES-256-GCM encrypt → upload to Blossom → kind-15 rumor). Keeps the existing privacy model (opaque ciphertext blob, no provider URL in the wire message) at the cost of a round-trip (fetch from provider, then upload to Blossom) before send.
  - Recommend **Option B** for consistency with the project's existing encrypted-attachment model and threat posture; Option A trades that away for latency savings only.
- **Stickers tab** (visible in the reference screenshot) is out of scope for this issue unless explicitly requested — file as a follow-up if wanted.

## Acceptance criteria

- [ ] New `GIFs` tab/panel in the composer, reachable from the same toolbar area as the emoji button, on both DM (`MessageInput.svelte`) and squad channel surfaces that reuse it.
- [ ] Search-as-you-type against the chosen provider, plus a trending/recent default view.
- [ ] Tapping a result sends the GIF as a message (encrypted attachment per Option B above), with the same delivery/receipt behavior as an image attachment (blurhash placeholder, decrypt-on-receive, etc. per `docs/messaging/ATTACHMENTS.md` §5).
- [ ] Provider API key lives server-side (Rust `std::env`), never in `VITE_*`/`ALCHEMY_*` or the webview bundle; new proxy command is registered in `src-tauri/src/lib.rs` and actually called from the frontend (not orphaned — see `pnpm check:tauri-commands`).
- [ ] New strings added to `en` and `es` `messaging.json` catalogs; no hardcoded user-facing text.
- [ ] `docs/messaging/ATTACHMENTS.md` (or a new doc) updated to disclose the new third-party network call and its metadata exposure, consistent with the existing Blossom disclosure.
- [ ] Manual verification via the Tauri MCP bridge in `make dev-sandbox` (per this repo's UI validation workflow): open picker, search, send a GIF in a DM, confirm it renders as a media tile for both sender and receiver.

## References

- `src/components/dm/MessageInput.svelte` — composer, emoji panel, attachment menu
- `docs/messaging/ATTACHMENTS.md` — existing encrypted attachment pipeline and its privacy disclosures
- `src/lib/i18n/locales/{en,es}/messaging.json` — translation catalogs

logicminds:
## Summary

Adds squad sticker packs and an opt-in GIF picker (via Klipy) to the DM/channel composer (#200), plus unblocks local two-account testing.

Issue #200 asked for a Telegram-style GIF picker and recommended fetching, encrypting, and re-uploading each GIF through the existing attachment pipeline. Klipy's terms forbid exactly that — media must be served from the URLs Klipy returns, never re-hosted — so that design was ruled out during implementation. The composer now ships two separate media layers instead of one: sticker packs (private, MLS-native) and GIFs (third-party, consent-gated).

## Design decisions

- **Sticker packs carry no new trust boundary.** A pack is a typed `sticker_pack_updated` announce over MLS, mirroring the existing `governance_updated` pattern — authorization is group membership, not a keypair or anything visible to a public relay. Images reuse the existing encrypted attachment pipeline (zero new wire format, zero new receive code). Packs converge last-write-wins on `updated_at` so out-of-order relay delivery can't resurrect a deleted pack.
- **GIFs are the first attachment Pacto sends unencrypted.** Sending a GIF carries the provider URL rather than bytes. Two boundaries replace encryption as the safety guarantee: the search tab stays inert behind a per-account disclosure the user must accept, and every Klipy media fetch is allowlisted to Klipy's CDN hosts in Rust — checked on each redirect hop, size-capped, and content-type-checked — so a peer can't use a crafted attachment to make the app fetch an arbitrary address. The disclosure gates *search*, not *rendering*: a recipient's client still fetches a received GIF from Klipy to display it, exposing that recipient's IP even if they never opened the tab. The disclosure copy says so explicitly, and `docs/messaging/GIF_PROVIDER.md` §3 documents that there is no recipient-side opt-out today.
- **API key stays server-side.** All Klipy egress goes through one Rust module (`src-tauri/src/klipy.rs`), keeping the key out of the webview and leaving a single chokepoint for the Tor routing tracked in #173.
- **Hardened decrypt path.** `crypto::decrypt_data` validates decoded key and nonce lengths and returns an error instead of panicking in `GenericArray::from_slice`, so an attacker-supplied `key`/`nonce` from an MLS announce can't crash a Tauri command. `fetch_sticker_image` rejects malformed params before issuing any download, and announce ingest caps a pack at 300 entries / 128 KiB.

## Also in this PR

`make dev-buddy` previously started a second `pnpm tauri dev`, which fought the first over port 1420 and never produced a usable second window — there was no way to exercise two-account behavior (squad invites, MLS round-trips) locally. It now reuses the debug binary and the running Vite server, and sandboxed runs skip the single-instance plugin. Four migration tests also stopped hardcoding version `32` as "newest" and now derive it from `embedded_ceiling()`/`embedded_migration_set()`, so they no longer break on every new migration.

## Validation

Verified against the live Klipy API: no provider connection exists until the disclosure is accepted, search and send work end to end, and a sent GIF renders from memory without ever exposing its URL as message text.

Fixes #200
Related: #173

Screenshot 2026-08-14 at 6 00 57 PM
Screenshot 2026-08-14 at 6 00 43 PM


src/components/dm/MessageInput.svelte
```diff
@@ -898,10 +1085,73 @@
...
+                      <button
+                        type="button"
+                        class="emoji-picker-item gif-picker-item"
+                        role="gridcell"
+                        disabled={disabled || isSendingAttachment}
+                        aria-label={gif.title || gif.slug}
+                        title={gif.title}
+                        on:click={() => selectGif(gif)}
+                      >
+                        <img src={gif.previewUrl} alt={gif.title} loading="lazy" />
```
daopunk: **Blocker — picker thumbs bypass the Rust chokepoint.**

This `` is a webview fetch. The PR and `docs/messaging/GIF_PROVIDER.md` say every Klipy byte request goes through `klipy.rs`, never the webview. Received GIFs honor that (`klipy_fetch_media` → blob URL). Search thumbs do not.

`previewUrl` is also passed through from Klipy JSON with no host allowlist in `map_item`. CSP is already `null`, so the webview will fetch whatever host the payload carries.

Render these the same way as received GIFs (allowlisted `klipy_fetch_media` → blob URL), or drop any URL whose host is not in `KLIPY_MEDIA_HOSTS` before the DTO leaves Rust — and still do not put those URLs in ``.

src-tauri/src/klipy.rs
```diff
@@ -0,0 +1,659 @@
...
+    file.hd
+        .as_ref()
+        .and_then(|s| s.gif.clone())
+        .or_else(|| file.md.as_ref().and_then(|s| s.gif.clone()))
+        .or_else(|| file.sm.as_ref().and_then(|s| s.gif.clone()))
+}
+
+/// Drops an item outright rather than surfacing a half-populated result if
+/// neither fallback chain finds a usable `gif` variant.
+fn map_item(item: RawItem) -> Option<KlipyGifDto> {
```
daopunk, line 192: **Blocker — URLs leave Rust unvalidated.**

`pick_preview` / `pick_full` copy Klipy’s `url` byte-identical. Nothing here checks `is_klipy_media_url`. Combined with the picker ``, a malformed or hostile payload becomes a webview request to an arbitrary host.

Filter both URLs through `is_klipy_media_url` in `map_item` and drop the item if either fails. That is cheap even if thumbs later move onto `klipy_fetch_media`.

src-tauri/src/klipy.rs
```diff
@@ -0,0 +1,659 @@
...
+}
+
+/// Fetches Klipy-hosted media bytes for rendering a received GIF attachment.
+/// This is the only place a Klipy media byte is ever fetched from: never the
+/// webview, never a generic downloader, and the bytes are handed back for an
+/// in-memory render only — nothing here ever touches disk, per Klipy's
+/// no-retain terms. The host allowlist keeps an attachment that merely claims
+/// a "Klipy" URL from turning this into an open fetch/SSRF primitive.
+#[tauri::command]
+pub async fn klipy_fetch_media(url: String) -> Result<Vec<u8>, String> {
```
daopunk: **Blocker — unbounded allowlisted GET.**

`HTTP_CLIENT` (line 30) uses reqwest’s default redirect policy, and the host allowlist is checked only on the original URL. An open redirect on a Klipy CDN host bypasses the SSRF tests below.

There is also no `Content-Length` / max-bytes cap and no content-type check. Recipients auto-fetch on render, so a peer can point a “GIF” at a large object on an allowlisted host.

`Vec<u8>` over JSON IPC becomes `number[]` in the webview — a few-MB GIF is a multi-million-element array.

Please: `redirect(Policy::none())` (or re-check host after each hop), cap the body, reject the wrong content-type, and return bytes in a Tauri-friendly encoding (base64 or a byte response), not a raw `Vec<u8>`.

src-tauri/src/message.rs
```diff
@@ -1271,6 +1271,224 @@ pub async fn message(
...
+    receiver: String,
+    url: String,
+    slug: String,
+    replied_to: String,
+) -> Result<bool, String> {
+    crate::session::heartbeat();
+    let handle = TAURI_APP.get().ok_or("App handle not available")?;
+    crate::migration::require_key_derivation_version_2_on_handle(handle)?;
+
+    if url.trim().is_empty() {
```
daopunk: **Blocker — send path does not enforce a Klipy URL.**

Empty string is the only check. Any `invoke` can send `https://evil.example/x` as an empty-key kind-15. Recipients will not fetch non-Klipy hosts (good), but this command should refuse the same way `klipy_fetch_media` does. Right now the boundary is “hope the picker is the only caller.”

src-tauri/src/sticker_pack.rs
```diff
@@ -0,0 +1,444 @@
...
+
+    let encrypted = crate::net::download(
+        &url,
+        &handle,
+        &cache_key,
+        Some(std::time::Duration::from_secs(30)),
+    )
+    .await
+    .map_err(|e| e.to_string())?;
+    let decrypted = crypto::decrypt_data(&encrypted, &key, &nonce)?;
```
daopunk, line 225: **Blocker — decrypt still panics; this path is attacker-reachable.**

The PR says the decrypt path now rejects empty keys. `crypto::decrypt_data` still does `GenericArray::from_slice` on the decoded key/nonce. The new guards are only on `download_attachment` / `decrypt_and_save_attachment`.

This command takes `key`/`nonce` from the MLS announce, downloads first (`net::download` — no host allowlist), then decrypts. A member can publish `key: ""` or `key: "aa"`. Opening the Stickers tab prefetches every entry (`MessageInput.svelte` ~690), so that is a panic in a Tauri command plus an arbitrary GET.

Validate key/nonce length in `decrypt_data` (return `Err`, never `from_slice` on the wrong size), and do the same check here *before* download. Cap pack entry count/bytes on ingest too.

src/components/dm/MessageAttachment.svelte
```diff
@@ -41,9 +42,50 @@
...
+   * through the Rust egress chokepoint only when the URL's host is on
+   * Klipy's CDN allowlist; the authoritative check lives in Rust
+   * (`klipy_fetch_media`), this is a client-side convenience gate. */
+  $: isKlipyGif = attachment.key === '' && attachment.nonce === '' && isKlipyMediaUrl(attachment.url);
+
+  let klipyBlobSrc: string | undefined;
+  let klipyUnavailable = false;
+  let klipyFetchedForUrl: string | undefined;
+
+  $: if (isKlipyGif && attachment.url !== klipyFetchedForUrl) {
```
daopunk, line 56: **Should-fix — receive path is not consent-gated.**

This reactive block auto-fetches Klipy as soon as the message is on screen. `fetchGifMedia` does not call `assertGifsDisclosureAccepted()`. A user who never opened the GIFs tab still sends their IP to Klipy.

`GIF_PROVIDER.md` §3 is honest about this. The PR body (“fail-closed — no Klipy request can leave the device before [disclosure]”) is not. At minimum, update the disclosure copy so senders know the recipient’s IP is exposed; a recipient-side prompt or “load GIF” click would match the privacy posture better.

src/components/dm/MessageInput.svelte
```diff
@@ -630,8 +652,172 @@
...
+  const STICKER_MIME_EXTENSIONS: Record<string, string> = {
+    'image/png': 'png',
+    'image/jpeg': 'jpg',
+    'image/gif': 'gif',
+    'image/webp': 'webp',
+  };
+
+  let stickerImageCache: Record<string, string> = {};
+
+  async function ensureStickerImageCached(entry: StickerEntry) {
```
daopunk, line 695: **Should-fix — opening Stickers prefetches every pack URL.**

This downloads and decrypts every visible entry via `fetch_sticker_image` → `net::download`. No host allowlist, no entry cap. A member can announce a large `entries` array (or non-Blossom URLs) and make every other client fetch them the moment someone opens the tab.

Prefetch lazily (viewport / on click), and reject pack entries whose URL is not a known Blossom host.

src/lib/api/klipy.ts
```diff
@@ -0,0 +1,178 @@
...
+    const parsed = new URL(url);
+    return parsed.protocol === 'https:' && KLIPY_MEDIA_HOSTS[parsed.hostname] === true;
+  } catch {
+    return false;
+  }
+}
+
+/** Fetches Klipy media bytes through the Rust egress chokepoint for an in-memory
+ * render. Never persisted to disk (Klipy's no-retain terms). Backend: klipy_fetch_media. */
+export async function fetchGifMedia(url: string): Promise<Uint8Array> {
```
daopunk, line 172: `invoke<number[]>` is the frontend half of the `Vec<u8>` IPC problem. Even a medium GIF becomes a huge JSON array in the webview.

Also not gated on disclosure — intentional for receive, but it means the “fail-closed” claim is TS-search-wrappers only. The Rust commands themselves have no consent store.

src/components/dm/GifDisclosure.svelte
```diff
@@ -0,0 +1,77 @@
...
+    padding: 8px 10px;
+    border-radius: 8px;
+    font-size: 0.8125rem;
+    font-weight: 500;
+    cursor: pointer;
+  }
+
+  .gif-disclosure-accept {
+    border: none;
+    background: var(--accent);
```
daopunk: Theme contract: brand fills use `background: var(--brand)` and `color: var(--on-brand)` — no `--accent`, no `--bg-primary` as text-on-fill. Hover surfaces are `--bg-hover`; this is a primary action, so `--brand` / `--on-brand`.

daopunk (changes requested):
## Verdict: request changes

The product split is right — private MLS sticker packs vs a consent-gated Klipy GIF picker — and the write-up is strong. Several security claims in that write-up are not true of the code. I would not merge until the four blockers below are fixed.

**What works:** sticker announce ingest (MLS membership + `squad_id == chat`, LWW + tombstones), API key stays in Rust with redaction, `content_filter=high` / no `customer_id`, empty-key attachments refused on the generic download path, sticker cache keyed by URL+key, image sniffing on upload, i18n + Klipy attribution, `make dev-buddy`, and real tests for the disclosure gate and host allowlist.

---

### Blockers

1. **Picker thumbnails bypass the Rust chokepoint.** `` fetches Klipy (or whatever host the API returned) from the webview. `parse_klipy_page` does not allowlist URLs. CSP is `null`. Received GIFs go through `klipy_fetch_media`; search thumbs do not.
2. **`crypto::decrypt_data` still panics on short/empty keys.** Call-site guards were added on the generic download path only. `fetch_sticker_image` decrypts MLS-attacker-controlled `key`/`nonce`. Opening the Stickers tab prefetches every entry.
3. **`klipy_fetch_media` is an unbounded allowlisted GET.** reqwest follows redirects (allowlist is only on the original URL), no max body size, and `Vec<u8>` crosses IPC as a JSON number array.
4. **`klipy_gif_message` does not enforce a Klipy URL.** Empty string is the only check. The send command should refuse the same way `klipy_fetch_media` does.

### Should-fix

- Consent is TS-only. Recipients auto-fetch Klipy on render with no opt-out. The design doc is honest about this; the PR body (“no Klipy request can leave the device before disclosure”) is not. Disclosure copy should mention recipient IP too.
- Opening Stickers runs `net::download` for every pack URL with no host allowlist and no entry/size cap.
- `updated_at` is unix seconds; same-second announces are no-ops (`>`).
- `GifDisclosure` uses `--accent` / `--bg-primary` instead of `--brand` / `--on-brand`.
- Comments cite `#173` / `KD8` — tracker IDs do not belong in `src/` / `src-tauri/`.
- Scope: two-account Makefile, stickers, and Klipy in one PR. GIF/sticker UI was dumped into legacy `MessageInput.svelte` (~300 lines) instead of a runes child.

### Suggested merge bar

Fix blockers 1–4. The sticker half is closer to ready than the GIF half — splitting stickers + `dev-buddy` first would shrink the review surface.

src-tauri/src/klipy.rs
logicminds: > Filter both URLs through `is_klipy_media_url` in `map_item` and drop the item if either fails.
> `redirect(Policy::none())` (or re-check host after each hop), cap the body, reject the wrong content-type, and return bytes in a Tauri-friendly encoding
> this command should refuse the same way `klipy_fetch_media` does
> `invoke<number[]>` is the frontend half of the `Vec<u8>` IPC problem

All four confirmed and fixed together in 91b567c — they were one hole with four openings, so replying on each thread:

- **`map_item`** now drops any item whose picked preview *or* full URL fails `is_klipy_media_url`, so no unvalidated host leaves Rust. New fixture test covers a non-Klipy host being dropped from a parsed page.
- **`klipy_fetch_media`** now uses a dedicated `MEDIA_CLIENT` whose `redirect::Policy::custom` re-applies the allowlist to `attempt.url()` on **every hop** (refusing via `attempt.error(...)`) rather than only the original URL — I went with per-hop re-checking over `Policy::none()` so a legitimate CDN redirect inside the allowlist still works. It also pre-checks `Content-Length`, streams with a running total against a named `MAX_MEDIA_BYTES` (16 MiB), and requires an `image/` content-type. Return type is now `Result<tauri::ipc::Response, String>`, so bytes cross IPC as a raw `ArrayBuffer`.
- **`klipy_gif_message`** now calls the same `is_klipy_media_url` predicate (exported `pub(crate)`, not duplicated) and refuses a non-Klipy URL on send. The boundary no longer depends on the picker being the only caller.
- **`klipy.ts`** does `invoke<ArrayBuffer>` and returns `new Uint8Array(buf)` — no more `number[]`. Added `fetchGifBlobUrl()` as the single path by which Klipy media reaches a DOM `src`.

On your disclosure point: correct that the Rust commands have no consent store, and the receive path is deliberately not gated — I've left that thread open, since whether to add a recipient-side gate is a product call rather than a fix.

Verified live against the real Klipy API in a seeded sandbox: 24 trending results loaded and **all 24 thumbnails render from `blob:` URLs, zero `klipy.com` in any ``** — which also confirms the new redirect/size/content-type checks accept genuine Klipy media. `cargo test --lib` 732 passed, frontend 2037 passed.

Also dropped the `#173` / `KD8` tracker references from `src/` per the repo convention; that rationale stays in `docs/messaging/GIF_PROVIDER.md`.

src-tauri/src/klipy.rs
logicminds: > Filter both URLs through `is_klipy_media_url` in `map_item` and drop the item if either fails.
> `redirect(Policy::none())` (or re-check host after each hop), cap the body, reject the wrong content-type, and return bytes in a Tauri-friendly encoding
> this command should refuse the same way `klipy_fetch_media` does
> `invoke<number[]>` is the frontend half of the `Vec<u8>` IPC problem

All four confirmed and fixed together in 91b567c — they were one hole with four openings, so replying on each thread:

- **`map_item`** now drops any item whose picked preview *or* full URL fails `is_klipy_media_url`, so no unvalidated host leaves Rust. New fixture test covers a non-Klipy host being dropped from a parsed page.
- **`klipy_fetch_media`** now uses a dedicated `MEDIA_CLIENT` whose `redirect::Policy::custom` re-applies the allowlist to `attempt.url()` on **every hop** (refusing via `attempt.error(...)`) rather than only the original URL — I went with per-hop re-checking over `Policy::none()` so a legitimate CDN redirect inside the allowlist still works. It also pre-checks `Content-Length`, streams with a running total against a named `MAX_MEDIA_BYTES` (16 MiB), and requires an `image/` content-type. Return type is now `Result<tauri::ipc::Response, String>`, so bytes cross IPC as a raw `ArrayBuffer`.
- **`klipy_gif_message`** now calls the same `is_klipy_media_url` predicate (exported `pub(crate)`, not duplicated) and refuses a non-Klipy URL on send. The boundary no longer depends on the picker being the only caller.
- **`klipy.ts`** does `invoke<ArrayBuffer>` and returns `new Uint8Array(buf)` — no more `number[]`. Added `fetchGifBlobUrl()` as the single path by which Klipy media reaches a DOM `src`.

On your disclosure point: correct that the Rust commands have no consent store, and the receive path is deliberately not gated — I've left that thread open, since whether to add a recipient-side gate is a product call rather than a fix.

Verified live against the real Klipy API in a seeded sandbox: 24 trending results loaded and **all 24 thumbnails render from `blob:` URLs, zero `klipy.com` in any ``** — which also confirms the new redirect/size/content-type checks accept genuine Klipy media. `cargo test --lib` 732 passed, frontend 2037 passed.

Also dropped the `#173` / `KD8` tracker references from `src/` per the repo convention; that rationale stays in `docs/messaging/GIF_PROVIDER.md`.

src-tauri/src/message.rs
logicminds: > Filter both URLs through `is_klipy_media_url` in `map_item` and drop the item if either fails.
> `redirect(Policy::none())` (or re-check host after each hop), cap the body, reject the wrong content-type, and return bytes in a Tauri-friendly encoding
> this command should refuse the same way `klipy_fetch_media` does
> `invoke<number[]>` is the frontend half of the `Vec<u8>` IPC problem

All four confirmed and fixed together in 91b567c — they were one hole with four openings, so replying on each thread:

- **`map_item`** now drops any item whose picked preview *or* full URL fails `is_klipy_media_url`, so no unvalidated host leaves Rust. New fixture test covers a non-Klipy host being dropped from a parsed page.
- **`klipy_fetch_media`** now uses a dedicated `MEDIA_CLIENT` whose `redirect::Policy::custom` re-applies the allowlist to `attempt.url()` on **every hop** (refusing via `attempt.error(...)`) rather than only the original URL — I went with per-hop re-checking over `Policy::none()` so a legitimate CDN redirect inside the allowlist still works. It also pre-checks `Content-Length`, streams with a running total against a named `MAX_MEDIA_BYTES` (16 MiB), and requires an `image/` content-type. Return type is now `Result<tauri::ipc::Response, String>`, so bytes cross IPC as a raw `ArrayBuffer`.
- **`klipy_gif_message`** now calls the same `is_klipy_media_url` predicate (exported `pub(crate)`, not duplicated) and refuses a non-Klipy URL on send. The boundary no longer depends on the picker being the only caller.
- **`klipy.ts`** does `invoke<ArrayBuffer>` and returns `new Uint8Array(buf)` — no more `number[]`. Added `fetchGifBlobUrl()` as the single path by which Klipy media reaches a DOM `src`.

On your disclosure point: correct that the Rust commands have no consent store, and the receive path is deliberately not gated — I've left that thread open, since whether to add a recipient-side gate is a product call rather than a fix.

Verified live against the real Klipy API in a seeded sandbox: 24 trending results loaded and **all 24 thumbnails render from `blob:` URLs, zero `klipy.com` in any ``** — which also confirms the new redirect/size/content-type checks accept genuine Klipy media. `cargo test --lib` 732 passed, frontend 2037 passed.

Also dropped the `#173` / `KD8` tracker references from `src/` per the repo convention; that rationale stays in `docs/messaging/GIF_PROVIDER.md`.

src/lib/api/klipy.ts
logicminds: > Filter both URLs through `is_klipy_media_url` in `map_item` and drop the item if either fails.
> `redirect(Policy::none())` (or re-check host after each hop), cap the body, reject the wrong content-type, and return bytes in a Tauri-friendly encoding
> this command should refuse the same way `klipy_fetch_media` does
> `invoke<number[]>` is the frontend half of the `Vec<u8>` IPC problem

All four confirmed and fixed together in 91b567c — they were one hole with four openings, so replying on each thread:

- **`map_item`** now drops any item whose picked preview *or* full URL fails `is_klipy_media_url`, so no unvalidated host leaves Rust. New fixture test covers a non-Klipy host being dropped from a parsed page.
- **`klipy_fetch_media`** now uses a dedicated `MEDIA_CLIENT` whose `redirect::Policy::custom` re-applies the allowlist to `attempt.url()` on **every hop** (refusing via `attempt.error(...)`) rather than only the original URL — I went with per-hop re-checking over `Policy::none()` so a legitimate CDN redirect inside the allowlist still works. It also pre-checks `Content-Length`, streams with a running total against a named `MAX_MEDIA_BYTES` (16 MiB), and requires an `image/` content-type. Return type is now `Result<tauri::ipc::Response, String>`, so bytes cross IPC as a raw `ArrayBuffer`.
- **`klipy_gif_message`** now calls the same `is_klipy_media_url` predicate (exported `pub(crate)`, not duplicated) and refuses a non-Klipy URL on send. The boundary no longer depends on the picker being the only caller.
- **`klipy.ts`** does `invoke<ArrayBuffer>` and returns `new Uint8Array(buf)` — no more `number[]`. Added `fetchGifBlobUrl()` as the single path by which Klipy media reaches a DOM `src`.

On your disclosure point: correct that the Rust commands have no consent store, and the receive path is deliberately not gated — I've left that thread open, since whether to add a recipient-side gate is a product call rather than a fix.

Verified live against the real Klipy API in a seeded sandbox: 24 trending results loaded and **all 24 thumbnails render from `blob:` URLs, zero `klipy.com` in any ``** — which also confirms the new redirect/size/content-type checks accept genuine Klipy media. `cargo test --lib` 732 passed, frontend 2037 passed.

Also dropped the `#173` / `KD8` tracker references from `src/` per the repo convention; that rationale stays in `docs/messaging/GIF_PROVIDER.md`.

src/components/dm/MessageInput.svelte
logicminds: > Render these the same way as received GIFs (allowlisted `klipy_fetch_media` → blob URL) ... and still do not put those URLs in ``.
> Prefetch lazily (viewport / on click), and reject pack entries whose URL is not a known Blossom host.

Both fixed in 91b567c, replying on each thread since they landed together.

**GIF thumbnails.** `` is gone. Tiles now render a blob URL from `fetchGifBlobUrl()` → `klipy_fetch_media`, the same route received GIFs take, so the webview never touches Klipy. Fetches respect the existing `gifsRequestToken` so a resolve arriving after a new search or tab switch is discarded and its blob revoked; every cached blob URL is revoked on unmount, panel close, tab switch, and before a fresh results page replaces the list.

**Sticker prefetch — addressed differently.** The eager all-entries loop is gone; each tile carries a `use:stickerVisible` `IntersectionObserver` rooted at the picker's scroll body that fetches once on first intersection then unobserves. `insertSticker` still fetches on click.

**Declined: the Blossom host allowlist.** Generic Pacto attachments already accept any Blossom host by design, so gating pack entries to a fixed list would reject legitimate packs hosted elsewhere without closing anything the other two mitigations don't. Those are (1) lazy fetching here and (2) a 300-entry / 128 KiB cap on announce ingest in `db.rs`. Together they remove the actual attack — involuntary mass fetch fan-out on tab open — without narrowing where a legitimate pack may live. Happy to revisit if you think a legitimate pack should be host-restricted.

Verified live in a seeded sandbox with a 300-entry pack, panel showing 274px of 1486px of content:

- On open: **only the on-screen tile fetched.** Tile 299 was pre-seeded into the sticker cache so it would render instantly if requested — it stayed blank, proving off-screen entries are never fetched.
- After scrolling to the bottom: tile 299 rendered.

Before this change that same pack fired 300 `fetch_sticker_image` calls the instant the tab opened. Separately, all 24 GIF thumbnails rendered from `blob:` URLs with no `klipy.com` in any ``.

Also dropped the `KD8` reference from the `sendGifUrl` comment per the no-tracker-IDs-in-`src/` rule.

src/components/dm/MessageInput.svelte
logicminds: > Render these the same way as received GIFs (allowlisted `klipy_fetch_media` → blob URL) ... and still do not put those URLs in ``.
> Prefetch lazily (viewport / on click), and reject pack entries whose URL is not a known Blossom host.

Both fixed in 91b567c, replying on each thread since they landed together.

**GIF thumbnails.** `` is gone. Tiles now render a blob URL from `fetchGifBlobUrl()` → `klipy_fetch_media`, the same route received GIFs take, so the webview never touches Klipy. Fetches respect the existing `gifsRequestToken` so a resolve arriving after a new search or tab switch is discarded and its blob revoked; every cached blob URL is revoked on unmount, panel close, tab switch, and before a fresh results page replaces the list.

**Sticker prefetch — addressed differently.** The eager all-entries loop is gone; each tile carries a `use:stickerVisible` `IntersectionObserver` rooted at the picker's scroll body that fetches once on first intersection then unobserves. `insertSticker` still fetches on click.

**Declined: the Blossom host allowlist.** Generic Pacto attachments already accept any Blossom host by design, so gating pack entries to a fixed list would reject legitimate packs hosted elsewhere without closing anything the other two mitigations don't. Those are (1) lazy fetching here and (2) a 300-entry / 128 KiB cap on announce ingest in `db.rs`. Together they remove the actual attack — involuntary mass fetch fan-out on tab open — without narrowing where a legitimate pack may live. Happy to revisit if you think a legitimate pack should be host-restricted.

Verified live in a seeded sandbox with a 300-entry pack, panel showing 274px of 1486px of content:

- On open: **only the on-screen tile fetched.** Tile 299 was pre-seeded into the sticker cache so it would render instantly if requested — it stayed blank, proving off-screen entries are never fetched.
- After scrolling to the bottom: tile 299 rendered.

Before this change that same pack fired 300 `fetch_sticker_image` calls the instant the tab opened. Separately, all 24 GIF thumbnails rendered from `blob:` URLs with no `klipy.com` in any ``.

Also dropped the `KD8` reference from the `sendGifUrl` comment per the no-tracker-IDs-in-`src/` rule.

src-tauri/src/sticker_pack.rs
logicminds: > Validate key/nonce length in `decrypt_data` (return `Err`, never `from_slice` on the wrong size), and do the same check here *before* download. Cap pack entry count/bytes on ingest too.

Right on all three, and the diagnosis was exact — the PR body's "hardened decrypt path" claim described call-site guards on `download_attachment` / `decrypt_and_save_attachment`, not the primitive. Fixed in 91b567c:

- **`crypto::decrypt_data`** now validates the decoded key (32 bytes) and nonce (16 bytes) immediately after `hex::decode` and returns a descriptive `Err`, so `GenericArray::from_slice` is never reached with a wrong-sized slice. That closes it for every caller, not just the sticker path. Tests assert `Err` for an empty key, `"aa"`, and a short nonce, plus a positive round-trip so the guard can't over-reject.
- **`fetch_sticker_image`** calls a new `validate_sticker_crypto_params` as its first statement, before `sticker_cache_dir` or `net::download` — a malformed announce entry can no longer make the client issue a GET at all, which was the arbitrary-fetch half of the report.
- **Ingest caps**: an announce is dropped as a no-op when `entries` exceeds 300 items or 128 KiB serialized.

I also fixed the `updated_at` second-precision issue from your should-fix list here, since it lives in the same files: it now stamps milliseconds. The strict `>` comparison stays strict — millisecond precision is what makes two edits in the same second distinct, whereas relaxing to `>=` would let a stale tombstone overwrite an equal-timestamp edit. Documented in `STICKER_PACKS.md` along with the caps.

`cargo test --lib` 732 passed. Verified live that a pack DTO now carries a 13-digit `updatedAt`.

src/components/dm/GifDisclosure.svelte
logicminds: > Theme contract: brand fills use `background: var(--brand)` and `color: var(--on-brand)` — no `--accent`, no `--bg-primary` as text-on-fill.

Fixed in 91b567c — `.gif-disclosure-accept` is now `background: var(--brand)` / `color: var(--on-brand)`. Audited the rest of the style block: `--text-primary`, `--text-muted`, and `--border-subtle` on the title and decline button were already on-contract, and the accept button had no hover state so none was added.

Verified live on `dark-techno`: computed `background-color` resolves to `rgb(34, 211, 238)` (`--brand` `#22d3ee`) and `color` to `rgb(12, 18, 25)` (`--on-brand` `#0c1219`).

src/components/dm/MessageAttachment.svelte
logicminds: > At minimum, update the disclosure copy so senders know the recipient's IP is exposed; a recipient-side prompt or "load GIF" click would match the privacy posture better.

You're right that the PR body oversold this, and I've corrected it — the "fail-closed" claim described the search wrappers, not the receive path.

Landed the minimum in 91b567c: the sender-side disclosure now names the second exposure explicitly, in both `en` and `es` — that when they send a GIF, the recipient's client fetches it from Klipy to render it, so the recipient's IP reaches Klipy too, even if that recipient never opened the tab or accepted anything. `GIF_PROVIDER.md` §3 keeps its honest statement that there is no recipient-side opt-out, and now records that the UI says so too.

Leaving this thread open for the recipient-side gate itself. §3 documents auto-render as a deliberate choice rather than an oversight, and swapping it for click-to-load is a real product trade — it costs the inline-GIF experience in every thread, and a "load GIF" button arguably teaches users that GIFs are dangerous in a way plain images (which already fetch from arbitrary Blossom hosts on render) don't. That's a call I'd rather you make than have me quietly reverse. My lean is a per-account "auto-load GIFs from senders" preference defaulting to on, so the privacy-maximalist path exists without penalizing everyone — but it's your call, and it's genuinely separable from the four blockers.

logicminds:
> I would not merge until the four blockers below are fixed.

All four blockers are fixed in 91b567c, each with a reply on its own thread. Summary, plus the should-fix items that had no inline thread:

**Blockers**

1. **Picker thumbnails** no longer use `` — they render blob URLs through `klipy_fetch_media`, and `map_item` drops any item whose preview or full URL fails the CDN allowlist, so an unvalidated host never leaves Rust.
2. **`crypto::decrypt_data`** validates decoded key/nonce lengths and errors instead of reaching `GenericArray::from_slice`. `fetch_sticker_image` rejects malformed params *before* `net::download`, so a hostile announce can't even trigger the GET.
3. **`klipy_fetch_media`** re-checks the allowlist on every redirect hop, caps the body at 16 MiB, requires an `image/` content-type, and returns `tauri::ipc::Response` so bytes cross IPC as an `ArrayBuffer` instead of a JSON number array.
4. **`klipy_gif_message`** refuses a non-Klipy URL using the same predicate.

**Should-fix**

- **Consent on receive** — took your stated minimum: the disclosure copy now names the recipient-IP exposure in both locales, and I corrected the PR body, which was wrong. Left that thread **open**: swapping auto-render for click-to-load is a product trade, and §3 documents auto-render as a choice rather than an oversight. My lean is a per-account preference defaulting to on. Your call.
- **Sticker prefetch** — now lazy via `IntersectionObserver`, plus a 300-entry / 128 KiB ingest cap. I declined the Blossom host allowlist; reasoning on that thread.
- **`updated_at` seconds** — now milliseconds, keeping the strict `>` comparison (ms precision is what makes same-second edits distinct; `>=` would let a stale tombstone win a tie).
- **Tracker IDs** — `#173` and `KD8` removed from `src/` and `src-tauri/`; rationale stays in `docs/messaging/GIF_PROVIDER.md`.
- **Scope** — not addressing in this PR; see below.

**Validation:** `cargo test --lib` 732 passed, frontend 2037 passed, `pnpm check` and `pnpm lint` clean. Verified live in a seeded sandbox against the real Klipy API: 24 trending thumbnails all rendered from `blob:` URLs with zero `klipy.com` in any ``; and with a 300-entry pack (274px of 1486px visible) only the on-screen tile fetched on open, while a pre-cached tile 299 stayed blank until scrolled into view — that same pack previously fired 300 `fetch_sticker_image` calls on tab open.

**On splitting the PR:** I'd rather you decide than split it unilaterally. The four blockers are fixed here, so the security argument for splitting is weaker now, and the sticker/`dev-buddy`/Klipy slices are interleaved in `MessageInput.svelte` and `lib.rs` — untangling them means rewriting history on a branch you've already reviewed once. Same for extracting the picker UI into a runes child: it's the right end state per the repo's Svelte 5 direction, but it's a ~300-line move out of a legacy composer with no component-test coverage to catch a regression, which is a poor thing to bundle into a review-fix pass. Happy to do either — as its own PR, or here if you'd prefer.

daopunk (approved):

PR #244, opened 2026-08-11
Boot agent sandboxes authenticated, local-routed, and collision-free

logicminds:
## What this unlocks

Every fresh branch used to boot an agent into an empty account-creation screen on fixed ports, talking to production relays, behind a twelve-keypress PIN ritual. This lands Wave 1 of [the parallel-agent sandbox plan](docs/plans/2026-08-10-001-feat-parallel-agent-sandboxes-plan.md) — the four pacto-app primitives a one-command dev world is built on. After this, an agent's **first** DOM snapshot shows the authenticated navbar, two worktrees can run `make dev` at the same time without colliding, and a sandbox can be pointed at a local relay stack that a release build is structurally incapable of reaching.

Epic `pacto-app-384`; closes U2a–U2f, U3a–U3g, U4a–U4b, U5a–U5g (`pacto-app-384.2` … `.23`).

## The four primitives

**Runtime-resolved trusted relay set (U2).** The `TRUSTED_RELAYS` static is gone. All ~30 call sites read `trusted_relays::trusted_relays()`, and the list is private to that module, so a missed call site is now a compile error rather than a silent send to a production relay. Debug builds resolve the set once at startup from `PACTO_TRUSTED_RELAYS`; **the release accessor contains no environment read at all**, so a set variable cannot redirect a shipped build. A malformed entry, or an override that resolves to zero relays, refuses at boot instead of silently shortening the list or disabling messaging outright. When the override points at a local `wss://` endpoint that will not validate, the startup probe names the certificate problem rather than surfacing a connection error three layers down. **The `caddy trust` remediation it currently prints is wrong**, tracked as `pacto-app-384.69`: the relay websocket resolves through `async-wsocket`, which pins `tokio-tungstenite` to `rustls-tls-webpki-roots` — a hermetic trust store that never reads the OS keychain — so no host-side trust command can help. The real fix is a build flag that also enables `rustls-tls-native-roots`.

**Per-worktree ports and a sandbox handle (U3).** `scripts/dev-ports.mjs` derives a deterministic port set from the branch name (`main` keeps today's 1420/1421/9223 exactly), probes for occupancy, and advances on collision. The MCP bridge is part of that set — it was the easily-missed second port that left a parallel agent unable to drive its app at all — and now binds loopback only instead of `0.0.0.0`. Each sandbox writes `sandbox-handle.json` with the ports it *actually* bound, its root, its relay endpoints, and its identity, so an agent reads where to connect instead of assuming defaults.

**Zero-keystroke authenticated boot (U5).** One debug-gated `dev_login` command replaces `test_login_fixture`. Backend depth keeps the e2e harness's UI-less setup; full depth drives the existing recovery-phrase login path so the frontend session is indistinguishable from a human login. Two commands can no longer disagree about whether you are logged in — the documented MCP footgun this removes. A sandbox-only refusal mechanism is included but **inert in this PR**: it fires only when `PACTO_DEV_IDENTITY_SANDBOX_ONLY=1`, and nothing in the repo sets that today (the only references are the reader and a test teardown). Its caller arrives in Wave 2 (U7c), where the dev-world launcher exports it from the world manifest's `sandboxOnly` field. Until then, autologin with a mnemonic against the default trusted set is a live-network identity — treat it as such.

**dev-world data-directory refusal (U4).** `PACTO_DEV_WORLD` set without a sandbox root exits before any account or database work, enforced in Rust rather than in `make` so running the binary directly cannot bypass it. Logout's legacy MLS cleanup was the one remaining path that reached the real OS app-data directory from a sandboxed run; it now routes through the sandbox helper.

## Verified against a running app, not just tests

- **AE7** — `make dev-sandbox` with `PACTO_DEV_LOGIN_MNEMONIC` set: the first `webview_dom_snapshot` returned the authenticated navbar (Commons / DMs / Squads / Catch up), zero keyboard or click calls issued.
- **Handle** — recorded port index 1, bridge 9323 (the port the plugin bound), and the logged-in npub; the e2e harness connected to the handle-reported port on every run.
- **Relay override** — with `PACTO_TRUSTED_RELAYS=wss://localhost:7001` the handle listed only the local endpoint, and the probe emitted `invalid peer certificate: UnknownIssuer`. Wave-2 follow-up established the printed `caddy trust` remediation is not actionable (`384.69`); the diagnosis is right, the suggested cure is not.
- **Refusals** — `PACTO_DEV_WORLD=1` without a root, a malformed override, and an empty override each exited naming the specific problem; a sandbox-only identity against public relays was refused with the identity and all three relays named, leaving the app on the normal welcome screen with no half-set auth state.
- **Idempotence** — a second full-depth boot against the same sandbox root landed authenticated with no duplicate account.
- **Gates** — 628 `cargo test --lib`, 1870 Vitest, `pnpm check`, `pnpm lint`, `pnpm check:tauri-commands`, `cargo fmt --check`, and `make release-symbol-check` (now also forbidding `dev_login` and `PACTO_TRUSTED_RELAYS`) all green.

## Reviewer notes

- Contains commit `636f392` from #239 (per-branch dev data dirs), which this builds on — **#239 can be closed as superseded**.
- `cargo fmt` reformatted five files this PR does not otherwise touch (`db.rs`, `storage_format.rs`, three under `evm/`); that is pre-existing drift, and `cargo fmt --check` is clean for the first time as a result.
- Review fix applied on this branch (`969fd29`): `is_local_host` now matches the whole `127.0.0.0/8` loopback range and RFC 6761 `.localhost` aliases, not just `localhost`/`127.0.0.1`/`::1`. It gates the sandbox-only refusal, so the narrow match would have refused a valid loopback alias once that refusal has a caller.
- Follow-ups filed, not silently skipped:
  - `pacto-app-384.67` — the per-account *default user relay list* still adds public relays and does not honor the override. **Upgraded from "not fully closed" to blocking** by the Wave-2 probe: the gift-wrap subscription runs on the user relay pool, which never contains the trusted relay, so a sandbox pointed at a local relay cannot receive a welcome published only there. R2 is not partially done; it is not started.
  - `pacto-app-384.68` (P0) — `catch_up_entries` stores the gift-wrap event id while `accept_mls_welcome` looks it up in a store `mdk-core` keys by the rumor id, so a bot-originated invite can never be accepted. Blocks U7e.
  - `pacto-app-384.69` (P1) — see U2 above. Verified fixable via a `local-relay-tls` feature; kept opt-in so release relay trust is unchanged. Blocks U7c.
- The `tauri-e2e` job now passes. It never served the binary's compiled-in dev URL (a `cargo build` debug binary always has tauri's `dev` cfg set, so it loads `devUrl` and ignores `frontendDist`), and the sandbox handle was originally written inside Tauri's setup closure, so the harness hung waiting for it on a headless host. The harness serves that URL and the handle is now written before the app is built.
- Waves 2 and 3 (world-state manifest, dev-world orchestrator, reclaim, storage doctor, personas, scenarios, relay-free harness) are unstarted and mostly live in `pacto-dev-env`. U1, the invite-first probe, needs a live bot roster and is still open.


daopunk:
## Live `pnpm tauri:dev` path vs this PR

Huge +1 on Wave 1 — parallel sandboxes, ports, local relays, and authenticated boot are the right primitives for the agent/`make` path.

I want to flag a **second local testing path** that this PR does not cover, so we don’t close the related gate fix as “superseded.”

### How I test (and why not the sandbox yet)

I run **live clients on live networks**, not the bot/sandbox stack:

1. `pnpm tauri:dev` in one checkout
2. A second clone/folder with **manual** storage + port overrides
3. Two live clients on the same machine talking to real relays

I don’t use the sandbox as my primary env because I **need data** — real accounts, history, squads, live-network state — and that hasn’t been solved yet (world-state / personas / seeded scenarios are still later waves). Until the sandbox is robust enough that I can use it as my primary testing environment, I’m going to keep testing with live clients.

I am **not tied to landing [#246](https://github.com/covenant-gov/pacto-app/pull/246) as my PR**. I am tied to that live dual-client workflow still working after branch switches.

### Why #244 alone doesn’t unblock me

This PR correctly isolates data/ports for **`make dev` / `dev-sandbox` / fixtures** (and documents that as prevention for the storage-format brick). That is great for your path.

Bare **`pnpm tauri:dev` is unchanged** (`package.json` still just `tauri dev`). No per-branch data dir, no auto ports. So when I hop branches on a checkout that reuses the same profile storage, I still hit the launch gate when another branch left additive migration stamps — including PRs that never meant to own SQLite.

[#246](https://github.com/covenant-gov/pacto-app/pull/246) fixes that in the gate itself: hard-block only on **breaking** schema skew; auto-rewind **additive-ahead**. Your #244 `storage_format.rs` touch is fmt-only; the rewind logic is only in #246.

```text
#244  →  avoid collisions for make/sandbox/agent path
#246  →  don’t brick shared/custom storage on additive branch hops
         (including plain pnpm tauri:dev)
```

### Ask

Please **merge both**, or if you’d rather not take #246 as-is, we need **some** change that preserves live dual-`pnpm tauri:dev` testing across branch switches (gate softening and/or making isolation available without requiring `make`). Happy to reshape #246 or fold the gate fix into a follow-up — outcome that matters is the live path, not the PR number.

Happy to sync on merge order / conflict resolution (`OPERATOR_UPDATES.md` + `+layout.svelte` are the likely touch points).

daopunk:
## Technical review

**Verdict:** approve with notes — Wave 1 looks sound for the agent/`make` path. Release gating on `dev_login` / relay override is solid. A few debug-path footguns and one acknowledged R2 gap are worth tightening or documenting; none are “this must not merge” from a release-security perspective.

### What looks good

- Trusted relays: private list + accessor (missed site = compile error); release `init_from_env` has no env read; release-symbol check covers `dev_login` and `PACTO_TRUSTED_RELAYS`.
- `dev_login`: `#[cfg(debug_assertions)]` + exact `PACTO_ALLOW_TEST_AUTH=1`; unconfigured mnemonic is a clean skip.
- MCP bridge bound to `127.0.0.1` (was effectively wider before).
- `PACTO_DEV_WORLD` without sandbox root exits before account/DB work; path-escape tests look right.
- Logout legacy MLS cleanup goes through the sandbox helper.
- Handle written before app build + e2e serving `devUrl` — real harness fixes.
- Good unit coverage with env mutexes where needed.

### Findings

**Medium — sandbox-only refusal is opt-in.**  
`PACTO_DEV_IDENTITY_SANDBOX_ONLY` must be set for the “refuse recipe identity on non-local relays” path. `make dev-sandbox` sets `PACTO_ALLOW_TEST_AUTH=1` but not sandbox-only. So mnemonic + allow-test-auth against the default (production) trusted set can still full-login to public relays. PR description reads stronger than the default wiring. Prefer defaulting sandbox-only when a sandbox root is set, or document loudly that autologin without local relays / sandbox-only is a live-network identity.

**Medium (already filed) — `DEFAULT_RELAYS` still production.**  
Trusted override does not yet cover the per-account default user relay list (`pacto-app-384.67`). Wave 1 does not fully close “no traffic to production relays.” Keep that follow-up visible; don’t treat R2 as done.

**Low — `is_local_host` is loopback-only.**  
LAN / `*.localhost` aliases count as non-local. Fine if intentional; worth a one-liner in docs if operators expect private-LAN relays to count.

**Low — port probe TOCTOU.**  
Bind-check-then-later-claim in `dev-ports.mjs` can race under heavy parallel starts. Acceptable for local; failure mode is loud enough.

**Nit — fmt drive-by** on unrelated Rust files. Likely merge touch-points with open [#246](https://github.com/covenant-gov/pacto-app/pull/246) (`OPERATOR_UPDATES.md`, `+layout.svelte`, fmt’d `storage_format.rs`) if both land.

### Checked / not issues

- `adoptDevSession` using npub for `pubkey` matches existing `LoginKeyPair.public` bech32 usage.
- Autologin is `import.meta.env.DEV`-only; command stripped from release.
- Empty/malformed `PACTO_TRUSTED_RELAYS` refuses at boot.

Happy to discuss the sandbox-only defaulting change as a small follow-up on this branch or a fast Wave-1.1.

daopunk (approved):
approved with comments and ask

logicminds:
Went through all six findings against the code on the PR head. **All valid.** One is worse than reported, one has a remedy I think is wrong, and the workflow point is the one that actually changes the plan.

## Applied

**`is_local_host` (Low)** — fixed in `969fd29`. Widened to `is_loopback()` for both address families plus RFC 6761 `.localhost` aliases, so `127.0.0.2` and `relay.localhost` count as local. Slightly broader than you flagged: it isn't only a docs issue, because `all_relays_local()` gates the sandbox-only refusal, so once that refusal has a caller a valid loopback alias would have been refused with the relay named as if it were production. Two new tests (positive for the newly-local forms, negative for a LAN address and `notlocalhost.example.com`). 637 lib tests green, `cargo fmt --check` clean.

## Accepted, no change

**Port TOCTOU (Low)** — closing it means holding the listeners across the `make` → `pnpm` → `vite` process boundary, which isn't possible. Failure mode is a loud `EADDRINUSE` at startup, not silent corruption. Leaving it.

**fmt drive-by (Nit)** — acknowledged. `git merge-tree` against #246 gives exactly one conflict, `docs/build/OPERATOR_UPDATES.md`.

## Sharper than reported

**`DEFAULT_RELAYS` (Medium)** — you undersold this one. Probing it for Wave 2 showed the gift-wrap subscription (`lib.rs:3418`) runs on the *user* relay pool, which never contains the trusted relay. A sandbox pointed at a local relay **cannot receive a welcome published only there** — my probe only saw its welcome because the bot also publishes to a public relay. R2 isn't partially closed, it's structurally not started. `384.67` updated with the evidence and the PR description now says so.

**Sandbox-only (Medium)** — grepping the repo for `PACTO_DEV_IDENTITY_SANDBOX_ONLY` returns two hits: the reader and a test teardown. Nothing sets it. So it's not merely opt-in, it's **inert** — unreachable in every flow that exists. You're right that the description overstated it; that's now corrected.

**But I don't think your proposed fix is right.** Defaulting it on when a sandbox root is set breaks two real workflows:

- `AGENTS.md:261` and `TAURI_MCP_INTEGRATION.md:144` both prescribe `PACTO_DEV_LOGIN_MNEMONIC=... make dev-sandbox` with no relay override, so every agent UI verification would be refused at boot.
- It would refuse **your** dual-client setup the moment you use a mnemonic to skip the PIN.

The rule conflates two different things: `PACTO_TEST_SANDBOX_ROOT` says *where data lives*; sandbox-only says *where the key came from*. A human's random throwaway mnemonic isn't a recipe-derived key, and the app can't tell them apart — which is exactly why the signal has to come from the launcher. Its real caller is U7c: the dev-world launcher reads `sandboxOnly: true` from the world manifest and exports the flag. That field shipped in pacto-dev-env #19 and is schema-enforced there, so omitting it fails validation.

## Your workflow point — you're right, and it's worse than you argued

I checked whether any later wave rescues you. None does, by construction:

- **U15 is your PR**, but KTD5 scopes quarantine to *"only when a sandbox root is active — never against the real OS data directory."* Your profile **is** the real OS data directory. U15 skips you forever.
- **U6/U7 dev-world** answers "I need data" with bot-seeded local-relay data. You need live-network real accounts.
- **U3 ports** live in the Makefile. `pnpm tauri:dev` is bare `tauri dev`; `vite.config.ts:12` falls back to 1420. And `Makefile:51-52` sends `main` down the same unprotected path deliberately.

The plan's implicit bargain is isolation in exchange for synthetic data. You need real data, so you can't take the isolation, so nothing downstream reaches you. And you can't split the difference by copying a real profile into a sandbox root — two clients sharing MLS state on live relays is exactly the split-brain hazard KD8 forbids.

**So the concrete proposal: U15 gates on the wrong thing. `cfg(debug_assertions)`, not `sandbox_root().is_some()`.**

That fully answers my safety objection to #246 — every case I raised (reinstall, downgrade, rolled-back update, second machine) is by definition a **release** build, which keeps today's hard gate untouched — while covering bare `pnpm tauri:dev` and `make dev` on `main`. Same classifier, same move-aside-over-drop, same `sqlite_master.sql` comparison and WAL checkpoint; only the precondition changes. Noted on `384.56`.

Interim, no merge needed: `PACTO_TEST_SANDBOX_ROOT=... pnpm tauri:dev` works — it's a plain env read (`test_sandbox.rs:69`), not make-only. Buys isolation, not your live data, so it's a stopgap.

## Two things from the Wave-2 probe worth your time

**`384.68` (P0)** — `catch_up_entries` stores the gift-wrap event id, but `accept_mls_welcome` looks it up in a store `mdk-core` keys by the rumor id, so accepting returns `Welcome not found` and the squad never appears. Verified for bot-originated invites; **human-to-human invites are untested**. If you're doing live MLS work, worth a look from your side.

**`384.69` (P1)** — the `caddy trust` line the relay probe prints is not actionable. `async-wsocket` pins `tokio-tungstenite` to `rustls-tls-webpki-roots`, a hermetic trust store built from `RootCertStore::empty()` plus the compiled-in Mozilla list; `load_native_certs()` isn't in the binary, so no keychain is ever read. I hit this with the mkcert CA already installed and `openssl`/`curl` both validating the same endpoint. Fixable — `tokio-tungstenite` builds its root store additively, so a `local-relay-tls` feature enabling `rustls-tls-native-roots` makes it work — but deliberately opt-in, since enabling it by default would widen release relay trust to every CA the host trusts.

PR #246, opened 2026-08-11
fix: gate storage-format only on breaking schema skew

### LINKED ISSUE #245: fix: storage-format gate blocks branch switches when SQLite skew is additive-only
## Summary

The local **storage-format** launch gate (`This build cannot open your data`) treats any unrecognized `refinery_schema_history` stamp as a hard, machine-wide block. That makes everyday branch switching painful: once any checkout that embeds a newer migration unlocks a shared `io.pacto` profile, **every other checkout** without that migration file is bricked — including PRs that never touch SQLite.

This is **not** SemVer / app version. The UI copy (“newer version of Pacto”) is the wrong mental model for local git. The probe compares on-disk migration history to this checkout’s compile-time `embed_migrations!` set.

Remote `minimum_compatible_version` (intentional breaking releases) is out of scope and should stay as-is.

## Problem

1. Cold launch calls `get_storage_compatibility` → scans every `npub1*/pacto.db` → any `Unrecognized` / `Divergent` history → full-screen block for **all** accounts.
2. Dev and release share the same OS app-data dir (`identifier: io.pacto`, e.g. macOS `~/Library/Application Support/io.pacto/`).
3. Unlock on a migration-bearing branch stamps history; switching to an unrelated PR then fails the probe even when that PR has **zero** SQLite changes.
4. Today’s classifier only knows history version/name/checksum. It cannot tell `CREATE TABLE` (additive) from `ALTER` / destructive shape changes, and on the older checkout the newer SQL file is absent anyway.

Relevant code:

- `src-tauri/src/storage_format.rs` — classify / scan / `get_storage_compatibility`
- `src/lib/updater/update-gate.ts` — `resolveGateAtLaunch` blocks on `!allRecognized`
- `src/components/updater/UpdateGate.svelte` — storage-format copy
- `docs/build/OPERATOR_UPDATES.md` — recovery via manual quarantine

## Desired outcomes (acceptance)

| Case | Expectation |
|------|-------------|
| PR A and PR B: **no** SQLite changes | Switch A ↔ B with no gate |
| PR A: no SQLite / PR B: **adds** tables or indexes only (existing table shapes untouched) | Switch A ↔ B with no gate |
| PR A: no SQLite / PR B: **modifies** existing table shapes (migration) | Switching to A **is gated** (storage-format block) |

Notes:

- In this repo, “add tables” still ships as a refinery migration file (e.g. `CREATE TABLE …`). The distinction is **additive vs modifying**, not “migration file vs no file.”
- Additive rewind may **drop** B-only new tables when landing on A (old-table data kept). Returning to B re-runs B’s migration and recreates empty new tables — acceptable for greenfield additive ledgers.
- Treat **extra columns** on expected tables as additive (do not `DROP COLUMN`). **Missing** expected columns / divergent rewrite of a shared migration version = breaking → gate.

## Proposed approach

Keep the storage-format gate for **breaking** schema skew only. For **additive-ahead** skew, auto-rewind then continue launch.

```text
Launch probe each profile vs this build's expected schema
  → Recognized          → gate clear
  → AdditiveAhead       → backup DB, DROP extra tables/indexes, strip unrecognized history → clear
  → BreakingAhead       → storage-format hard gate (current UI)
```

### Classification (structural, not history-only)

1. Apply this build’s embedded migrations to an in-memory SQLite DB → **expected** tables, columns, indexes.
2. Compare to the real profile DB (`sqlite_master` + `PRAGMA table_info` for expected tables).
3. Verdicts:
   - **Recognized** — history matches embedded set (existing path).
   - **AdditiveAhead** — unrecognized history **and** every expected table has at least the expected columns **and** schema extras are only tables/indexes/columns beyond expected.
   - **BreakingAhead** — missing expected columns/objects, divergent name/checksum on a shared version, or otherwise unsafe structural diff.

### Launch / unlock policy

- `Recognized` → clear.
- `AdditiveAhead` → auto-rewind that profile (backup sibling e.g. `pacto.db.pre-rewind-<ts>`), re-probe, clear.
- Any `BreakingAhead` → machine-wide storage-format gate (intentional for true shape changes).
- Orphan cleanup: never `Delete` unrecognized / divergent / breaking profiles (keep current safety in `account_manager`).

### Likely touch points

- `src-tauri/src/storage_format.rs` — expected-schema snapshot; `AdditiveAhead` / `BreakingAhead`; additive rewind; richer compatibility report
- `src/lib/updater/update-gate.ts` — block only on breaking; invoke rewind for additive-ahead then clear
- Unlock / `get_db_connection` path — if still AdditiveAhead, rewind or clear error; BreakingAhead should already have gated
- Tests: additive `CREATE TABLE` + fake history → rewind → Recognized; missing expected column → BreakingAhead → gate
- `docs/build/OPERATOR_UPDATES.md` — document the three PR outcomes and additive rewind backups

## Out of scope

- Changing remote `minimum_compatible_version` behavior
- Branch-scoped sandboxes / forcing `PACTO_TEST_SANDBOX_ROOT` for `pnpm tauri:dev`
- Full down-migration files for historical `V1..V32`
- Preserving B-only rows in new tables while checked out on A

## Manual validation

1. Unlock a reused account on a branch that only **adds** a table → checkout a branch without that migration → `pnpm tauri:dev` opens and the same account unlocks (old data intact; new table dropped after backup).
2. Unlock on a branch that **alters** an existing table → checkout without that migration → storage-format gate still appears.
3. Two branches with no SQLite delta → switch freely with no gate and no rewind.

## Related

- Mandatory update gate design: `docs/plans/2026-08-05-001-feat-mandatory-update-gate-plan.md`
- Operator recovery today: `docs/build/OPERATOR_UPDATES.md` (“Recovering a machine the local storage-format check blocked”)

daopunk:
## Summary

Switching local git branches no longer bricks the whole app when a reused profile was last opened on a branch that only **added** SQLite tables/indexes/columns.

**Before:** Any unrecognized `refinery_schema_history` stamp triggered the machine-wide “This build cannot open your data” gate — including PRs that never touched SQLite, once some other checkout had stamped an additive migration on shared `io.pacto` data.

**After:** Launch compares on-disk schema to this build’s expected shape. Additive-ahead skew is auto-rewound (backup, drop extras, strip unrecognized history) and the app continues. The hard gate remains only for **breaking** skew (missing expected columns/objects, divergent checksum/name).

Closes #245.

## Changes

- Classify profiles as compatible / additive-ahead / breaking via structural schema compare (not history-only).
- Auto-rewind additive-ahead on the launch compatibility probe and before DB open.
- Keep the storage-format block UI for breaking profiles only; remote `minimum_compatible_version` unchanged.
- Document the three PR branch-switch outcomes in operator updates.

## Test plan

- [x] `cargo test --lib storage_format`
- [x] `pnpm exec vitest run src/lib/updater/update-gate.test.ts`
- [x] `pnpm check` / `pnpm lint`
- [ ] Manual: unlock on an additive-migration branch → checkout a branch without it → `pnpm tauri:dev` opens and the same account unlocks
- [ ] Manual: unlock on a branch that alters existing table shapes → checkout without it → storage-format gate still appears

---

### DIFF
--- a/docs/build/OPERATOR_UPDATES.md
+++ b/docs/build/OPERATOR_UPDATES.md
@@ -151,13 +151,21 @@ Use these in priority order.
 
    **After using either path to correct a value, mirror the final value back into the tracked `scripts/release-compatibility.json` and commit it.** The workflow and the local script both operate on the *published* `latest.json` only; they never write the tracked file. If the tracked file is left at the old value, the next tagged release re-runs the automatic stamping job with that stale value and silently regresses the correction.
 
-## Recovering a machine the local storage-format check blocked
+## Local storage-format check (breaking schema only)
 
-This is a **different** trigger from the minimum-version check above: it never touches the network. It fires when the app opens a profile's `pacto.db` on launch and finds a migration history that this build doesn't recognize — for example, a profile last opened by a newer or divergent build. Because the check runs before account selection, one bad profile blocks **every** account on that machine, not just the one it belongs to.
+This is a **different** trigger from the minimum-version check above: it never touches the network. At cold launch the app compares each profile’s on-disk schema to this build’s embedded migrations.
 
-The block screen deliberately does not show which npub or file path is at fault, so recovery is manual:
+**Branch / PR outcomes (shared `io.pacto` app data):**
 
-1. Note the schema version number the block screen reports.
+| Case | Result |
+|------|--------|
+| Neither branch changes SQLite | No gate |
+| Newer branch only **adds** tables/indexes/columns (existing shapes intact) | No gate — additive skew is auto-rewound (backup `pacto.db.pre-rewind-<ts>`, drop extras, strip unrecognized history) |
+| Newer branch **modifies** existing table shapes, or history diverges (checksum/name) | Storage-format hard gate |
+
+The block screen deliberately does not show which npub or file path is at fault. When the gate fires (breaking only), recovery is manual:
+
+1. Note the schema version number the block screen reports (when shown).
 2. Profile directories live under `<app_data_dir>/npub1…/pacto.db` — see [`../storage-layout/SQLITE_AND_FILES.md`](../storage-layout/SQLITE_AND_FILES.md) for the full on-disk layout. `<app_data_dir>` here is Tauri's per-OS app data directory *including* this app's identifier (`io.pacto`, from `src-tauri/tauri.conf.json`):
    - macOS: `~/Library/Application Support/io.pacto/`
    - Windows: `%APPDATA%\io.pacto\`
@@ -170,8 +178,10 @@ The block screen deliberately does not show which npub or file path is at fault,
 4. The one profile whose `version` matches the number from the block screen is the offender. Move that single directory aside (e.g. rename `npub1…` to `npub1…-quarantined`) — do not delete it.
 5. Relaunch the app. Every other account on the machine unblocks immediately; the moved-aside profile no longer participates in account discovery until it is restored (e.g. by a build that recognizes its schema).
 
+Additive auto-rewind keeps old-table data; new tables from the newer branch are dropped on the older checkout (a `.pre-rewind-*` sibling preserves the prior file). Returning to the newer branch re-runs its migration and recreates empty new tables.
+
 ## Limits of this gate
 
 - **The manifest is unsigned.** `latest.json`, including `minimum_compatible_version`, carries no signature — only the platform installers themselves are Ed25519-signed. Anyone who can write to the release's assets can alter the manifest.
 - **The gate is forward-only.** It cannot reach installs from before this feature shipped; those clients have no code path that reads `minimum_compatible_version`.
-- **The remote (minimum-version) trigger fails open by design**, so anyone able to interfere with the manifest fetch — blackholing it or corrupting the response — can suppress it entirely for a given client. The local storage-format trigger above is the one that survives this: it never depends on the network, so it cannot be defeated the same way.
+- **The remote (minimum-version) trigger fails open by design**, so anyone able to interfere with the manifest fetch — blackholing it or corrupting the response — can suppress it entirely for a given client. The local storage-format trigger above is the one that survives this for **breaking** schema skew: it never depends on the network. Additive-only skew is auto-rewound instead of blocking.
--- a/src-tauri/src/account_manager.rs
+++ b/src-tauri/src/account_manager.rs
@@ -506,6 +506,7 @@ pub fn get_db_connection<R: Runtime>(
 
     // Open new connection
     let db_path = get_database_path(handle, &npub)?;
+    crate::storage_format::prepare_profile_db_for_open(&db_path)?;
     let mut conn = rusqlite::Connection::open(&db_path)
         .map_err(|e| format!("Failed to open database: {}", e))?;
 
--- a/src-tauri/src/storage_format.rs
+++ b/src-tauri/src/storage_format.rs
@@ -1,14 +1,16 @@
-//! Read-only storage-format recognition for the app database (`pacto.db`).
+//! Storage-format recognition for the app database (`pacto.db`).
 //!
 //! Reproduces refinery-core 0.9.2's `verify_migrations` abort semantics
-//! ahead of time, on a read-only connection, so the app can tell a user to
-//! update instead of surfacing refinery's own opaque migration error. Never
-//! calls `run_migrations` or `get_db_connection` -- either would migrate the
-//! database as a side effect of merely checking it.
+//! ahead of time so the app can tell a user to update instead of surfacing
+//! refinery's opaque migration error. History-only `Unrecognized` stamps
+//! that are structurally additive (extra tables/indexes/columns only) are
+//! auto-rewound; shape-breaking skew still hard-blocks launch.
 
 use rusqlite::{Connection, OpenFlags};
-use std::path::Path;
-use std::time::{Duration, Instant};
+use std::collections::{HashMap, HashSet};
+use std::path::{Path, PathBuf};
+use std::sync::OnceLock;
+use std::time::{Duration, Instant, SystemTime, UNIX_EPOCH};
 use tauri::{AppHandle, Runtime};
 
 /// Busy-timeout mirroring `account_manager::account_has_valid_pkey`'s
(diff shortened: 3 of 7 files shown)


logicminds:
## Review: good classifier, but the rewind policy is too broad to merge as-is

Reviewed against `main@176b065` and against the sandbox work in #244. Short version: **the structural classifier is genuinely good and worth keeping. The auto-rewind policy is what blocks a merge**, because it changes release-build data-safety behaviour to fix a dev-workflow problem — and its own safety net has a silent bug.

Also: **you can unblock yourself right now without this PR.** Jump to the last section if that's all you need.

---

### 1. Destructive by default in release builds

`compatibility_report` runs on the cold-launch critical path for every build, and `prepare_profile_db_for_open` now runs inside `get_db_connection` (`account_manager.rs:509`). Neither is `cfg`-gated — there are **zero** `debug_assertions` in the 471 added lines of `storage_format.rs`.

So this isn't only a `git checkout` path. Any way a user ends up on an older build — reinstall, downgrade, a rolled-back auto-update, a second machine — takes the same `DROP TABLE`, silently, at launch, against their real account.

Concretely, using the commit that just landed on `main`: #242 added `squad_sponsored_fee_usage` (a pure additive `CREATE TABLE` + two indexes). Under this PR any build without that migration classifies the profile `AdditiveAhead` and drops the treasury sponsored-gas ledger. That is exactly the case the gate exists to prevent.

CI green doesn't speak to this — no test exercises a downgrade against a real profile.

### 2. The backup is taken before the WAL checkpoint

In `rewind_additive_profile`:

```rust
std::fs::copy(db_path, &backup)                          // ← copy
let conn = Connection::open(db_path)
conn.execute_batch("PRAGMA wal_checkpoint(TRUNCATE);")   // ← checkpoint
```

`get_db_connection` sets `journal_mode=WAL`, so `pacto.db-wal` normally holds committed frames. The copy grabs only `pacto.db` and neither `-wal` nor `-shm`. The resulting `pacto.db.pre-rewind-<ts>` looks intact and is missing recent commits — worse than no backup, because it is the thing we'd tell someone to restore from. (Restoring it next to a live mismatched `-wal` is its own hazard.)

Nothing prunes these either: every branch switch that trips the additive path leaves another full copy of the message DB on disk.

### 3. The classifier compares names, not definitions

`read_schema_snapshot` selects `type, name` from `sqlite_master` and column *names* from `PRAGMA table_info`. It never reads `sqlite_master.sql`. Consequences:

- a table rebuilt to add `NOT NULL` / `UNIQUE` or change a column type keeps its column names → **classified additive** → rewound → the older build runs against constraints it doesn't expect;
- an index redefined under the same name is invisible;
- triggers and views aren't in the snapshot at all, so a newer build's trigger **survives the rewind** and then fires against the rewound schema.

Comparing the `sql` text (and including `trigger`/`view`) closes all three.

### Minor

`prepare_profile_db_for_open` runs on every connection-pool *miss*, not just first open — and several of the ~100 `get_db_connection` call sites don't `return_db_connection` on error paths. That adds a read-only open plus a history query to hot paths like `get_pkey` and the session heartbeat.

### What's good and should survive

Building the expected schema by running `embed_migrations!` into an in-memory DB is the right trick and fits the existing style of `storage_format.rs`. Failing closed to history-only when that build fails is correct. Keeping a non-rewinding `scan_profiles` for other callers is thoughtful. The tests cover additive→clear, missing-column→breaking, divergent-checksum→breaking, and refuse-breaking-at-open.

---

## How this fits with #244 and the sandbox waves

Mechanically the two branches barely touch. `git merge-tree` against #244 gives exactly **one** conflict, `docs/build/OPERATOR_UPDATES.md` (we both add a branch-switch recovery section). `src/routes/+layout.svelte` auto-merges cleanly — your comment rewrite and our `void runDevAutologin()` land side by side. No overlap in `storage_format.rs`, `account_manager.rs`, or `src/lib/api/auth.ts`.

Architecturally there is a real collision, and it's the useful part of this review.

**#244 (Wave 1, open, all checks green) removes the cause of #245 rather than tolerating it.** It carries PR #239's commit — the one that gives plain `make dev` a per-branch data dir — which is *not* on `main` yet, and is why you're hitting this at all. On top of that it lands four debug-gated primitives:

- **U2** — trusted relay set resolved at runtime; debug builds can point all Nostr/MLS traffic at the local stack via `PACTO_TRUSTED_RELAYS`, release has no env read at all.
- **U3** — branch-hashed port set (Vite + HMR + MCP bridge) so N worktrees run concurrently, plus a `sandbox-handle.json` recording where each sandbox actually landed.
- **U4** — `PACTO_DEV_WORLD` refuses to start when no sandbox root resolved, so a world boot structurally cannot reach the real OS app-data directory.
- **U5** — one `dev_login` command; an agent's first DOM snapshot is already authenticated.

Net effect for this bug: after #244, two checkouts never share `io.pacto`, so an additive migration on branch B can't stamp branch A's profile in the first place.

**Wave 2** (`pacto-app-384.24`–`.41`) is the world itself, mostly in `pacto-dev-env`: the versioned world-state manifest and deterministic identity derivation (U6), the `dev-world` orchestrator with named readiness gates (U7), the delegating `make dev-world` alias (U8), local-chain artifact import with mismatch refusal (U9), and `make dev-world-reclaim` (U10).

**Wave 3** is where this PR belongs. **U15 — storage doctor and quarantine — is the same feature you've built here** (`pacto-app-384.56`, `.57`, `.58`), with one deliberate difference recorded as KTD5 in `docs/plans/2026-08-10-001-feat-parallel-agent-sandboxes-plan.md`:

> quarantine runs between profile scanning and the compatibility report, **and only when a sandbox root is active — never against the real OS data directory**. Quarantine *moves the offending profile directory aside* and records why.

Move-aside inside a sandbox root, versus `DROP TABLE` against whatever profile the launch scan finds. Merging this PR as-is would make U15 half-redundant and move the repo to the riskier posture. Wave 3 also brings U16 (shared-stack leases) and U17 (an adversarial fixture target that seeds exactly this kind of hostile state to prove the rails fire).

---

## Suggested path

1. **Unblock now, no code, works on `main` today:** `make dev-account` (or `dev-buddy` / `dev-sandbox`). Those targets already export `PACTO_TEST_SANDBOX_ROOT` (`Makefile:48,56,60`), the profile lands in `test_fixtures/`, and the storage-format scan resolves through `test_sandbox::test_data_dir` — so it never sees the other branch's stamp.
2. **Then merge #244**, which makes plain `make dev` do that per branch, automatically, for everyone.
3. **Retarget this PR onto U15** and keep the classifier:
   - gate `rewind_additive_profile` on `debug_assertions` **and** `test_sandbox::sandbox_root().is_some()`; release keeps today's hard gate;
   - checkpoint before copying (or copy `-wal`/`-shm` too), and prune old backups;
   - compare `sqlite_master.sql`, not just object names, and include triggers and views;
   - prefer moving the profile directory aside over dropping objects out of a live DB.

That keeps the good half of this work, closes #245 for the dev case immediately, and doesn't put a downgraded release user's data behind an automatic `DROP TABLE`.

Happy to open a branch that does the narrowing if it's easier than reworking this one.

daopunk:
@logicminds unclear whether this should be closed or not

daopunk:
yes, please open new narrower branch @logicminds

logicminds:
closing in favor of pre-seeded sandboxes

PR #258, opened 2026-08-12
build(ui): add shared tailwind v4 and shadcn-svelte base

thebeyondr:
## Summary

Pacto’s UI still runs mostly on plain CSS and ad-hoc styles. This PR adds a shared design-system foundation: **Tailwind CSS v4** (via `@tailwindcss/vite`) and a curated set of **shadcn-svelte** primitives under `$lib/components/ui`.

No product screens are redesigned here. The goal is one styling stack contributors can reuse instead of inventing CSS per feature.

## What changed

- Wired Tailwind v4 into Vite and imported it (plus `tw-animate-css`) in `src/app.css`
- Added `components.json` and `cn()` helpers in `src/lib/utils.ts`
- Checked in shadcn primitives: avatar, badge, button, dialog, dropdown-menu, input, popover, scroll-area, separator, tooltip
- Added deps: `bits-ui`, `@lucide/svelte`, `tailwind-merge`, `tailwind-variants`, etc. (pnpm lockfile only)
- Dialog “Close” copy goes through i18n (`ui.close` en/es)
- Small eslint override for the shadcn button’s href pattern
- Unit test for `cn()`

## Why

Without a shared utility + primitive layer, every UI change stays one-off CSS. Landing Tailwind + shadcn first gives consistent building blocks and a clear ownership path for UI primitives (treated as Pacto code, not throwaway CLI output).

## Test plan

- [ ] `pnpm install --frozen-lockfile`
- [ ] `pnpm check`
- [ ] `pnpm lint`
- [ ] `pnpm test`
- [ ] `pnpm build`
- [ ] Smoke existing screens still load (auth, settings, squads, DMs, dashboard) — Preflight is on; watch for reset regressions


src/app.css
```diff
@@ -1,3 +1,5 @@
+@import 'tailwindcss';
```
logicminds: **P0 (#1) -- Tailwind's `container` utility clamps the app shell on every screen.**

`<main class="container">` (`src/routes/+page.svelte:992`) is the app shell. This import turns `container` into a real utility: `width:100%` plus a `max-width` ladder at 40/48/64/80/96rem. Our own `.container` rule (`src/app.css:51-58`) declares neither `width` nor `max-width`, and a cascade layer only loses on properties the unlayered rule actually declares -- so Tailwind wins unopposed.

Measured on the production build of this branch, toggling only the Tailwind stylesheet, 1365px viewport:

| | baseline | this PR |
|---|---|---|
| `.container` width | 1349px | **1280px** |
| `.container` max-width | `none` | **1280px** |

Every window wider than 640px renders the shell clamped to the next breakpoint down, left-aligned, with dead `--bg-page` beside it -- auth, settings, squads, DMs and dashboard alike.

**Fix**
```css
.container {
  /* ...existing... */
  width: 100%;
  max-width: none; /* Tailwind ships a `container` utility; keep the shell full-bleed. */
}
```

logicminds: **P1 (#2) -- no `@theme` block, so every shadcn color utility compiles to nothing.** Found independently by 4 of 5 reviewers.

Tailwind v4 emits a color utility only when a matching `--color-*` theme variable exists. `shadcn-svelte init` normally writes that block into `app.css`; only the two `@import` lines came over. 25 distinct semantic utilities across the vendored set resolve to nothing.

Verified against the built CSS: the scanner *did* read the primitives -- `.inline-flex`, `.shrink-0`, `.size-8`, `.underline-offset-4` from `button.svelte` are all emitted -- yet `.bg-primary`, `.bg-popover`, `.text-popover-foreground`, `.border-ring`, `.bg-muted` appear **zero** times in any built asset. Rendered in-page, `class="bg-popover text-popover-foreground border-ring"` computes to `background: rgba(0,0,0,0)` with fully inherited color and border.

So: Dialog with no panel behind its text, Button with no fill, Input with no border, invisible Separator. Tailwind never errors on an unknown candidate, so build/check/lint/test all stay green.

**Fix** -- map onto the vars the five `[data-theme]` sheets already define, rather than pasting shadcn's neutral oklch palette (which would make primitives ignore the selected Pacto theme):
```css
@theme inline {
  --color-background: var(--bg-page);
  --color-foreground: var(--text-primary);
  --color-card: var(--bg-panel);
  --color-card-foreground: var(--text-primary);
  --color-popover: var(--bg-elevated);
  --color-popover-foreground: var(--text-primary);
  --color-primary: var(--accent);
  --color-primary-foreground: var(--bg-page);
  --color-secondary: var(--bg-panel);
  --color-secondary-foreground: var(--text-primary);
  --color-muted: var(--bg-hover);
  --color-muted-foreground: var(--text-muted);
  --color-accent: var(--bg-hover);
  --color-accent-foreground: var(--text-primary);
  --color-destructive: var(--danger);
  --color-border: var(--border);
  --color-input: var(--border-subtle);
  --color-ring: var(--accent);
}
```
`inline` keeps the `var()` reference live so per-theme switching still works.

src/app.css
```diff
@@ -1,3 +1,5 @@
...
+@import 'tw-animate-css';
```
logicminds, line 3: **P1 (#3) -- Preflight silently resets headings, block margins and list markers app-wide.** (Design call.)

Preflight wins on every property our hand-written CSS never declares. Our only global heading rule (`src/app.css:31-35`) sets font-family, weight and letter-spacing -- **not** font-size.

Measured on the production build, same page, toggling only the Tailwind stylesheet:

| | baseline | this PR |
|---|---|---|
| `<h2>` font-size | 24px | **16px** |
| `<h2>` margin-top | 19.92px | **0px** |
| `<ul>` list-style | `disc` | **`none`** |
| `<ul>` padding-inline-start | 40px | **0px** |

30 heading occurrences across Navbar, ChatView, UpdateGate, DeploySafeModal and the settings modals have no font-size rule anywhere and collapse to body text. Chat message bodies lose their bullet markers, and inline twemoji become block elements -- `FormattedMessageBody.svelte:212-216` styles `.twemoji` with height/width/vertical-align but no `display`, so Preflight's `img { display: block }` is the only declaration in play.

**Two defensible options** -- your call:
1. Import without Preflight (matches "no product screens are redesigned here"):
   ```css
   @import 'tailwindcss/theme.css' layer(theme);
   @import 'tailwindcss/utilities.css' layer(utilities);
   ```
2. Keep Preflight deliberately and restore the UA behaviours in this PR: heading font-sizes, `:global(.twemoji){display:inline}`, and list-style for message-body `ul`/`ol`.

The test-plan line "watch for reset regressions" is still unchecked -- this is what it would have caught.

src/lib/components/ui/avatar/avatar.svelte
```diff
@@ -0,0 +1,26 @@
...
+	} = $props();
+</script>
+
+<AvatarPrimitive.Root
+	bind:ref
+	bind:loadingStatus
+	data-slot="avatar"
+	data-size={size}
+	class={cn(
+		"size-8 rounded-full after:rounded-full data-[size=lg]:size-10 data-[size=sm]:size-6 group/avatar relative flex shrink-0 select-none after:absolute after:inset-0 after:border after:border-border after:mix-blend-darken dark:after:mix-blend-lighten",
```
logicminds, line 22: **P2 (#5) -- `dark:` keys off the OS color scheme, not our `[data-theme]` switch.** Found independently by 4 of 5 reviewers.

Tailwind's built-in `dark:` compiles to `@media (prefers-color-scheme: dark)` unless the stylesheet declares `@custom-variant dark`; `src/` declares none. We select among five themes via `document.documentElement.setAttribute('data-theme', ...)` (`src/stores/theme.ts:44`), and `techno` is a **light** theme (`--bg-page: #f0eeea`).

A user on `techno` with the OS in Dark gets dark refinements over a light surface; a user on `midnight` with a light OS never gets them. This one is independent of the missing tokens -- `dark:after:mix-blend-lighten` on this line references no custom color and already compiles under the wrong condition today.

**Fix** in `src/app.css`, beside the Tailwind import:
```css
@custom-variant dark (&:where([data-theme="dark-techno"] *, [data-theme="midnight"] *, [data-theme="aztec"] *, [data-theme="union"] *));
```
Assumption: `techno` is the only light theme -- adjust the list otherwise.

src/lib/utils.test.ts
```diff
@@ -0,0 +1,8 @@
+import { describe, expect, it } from 'vitest';
+import { cn } from './utils';
+
+describe('cn', () => {
+  it('combines conditional classes and keeps the last Tailwind conflict', () => {
```
logicminds, line 5: **P1 (#4) -- nothing in the toolchain can fail when a Tailwind utility compiles to nothing.**

This test asserts `cn('block px-2', { hidden: false }, ['px-4']) === 'block px-4'` -- which passes identically whether or not a single `--color-*` token exists. Meanwhile `pnpm build`, `pnpm check` (0 errors), `pnpm lint` (clean) and `pnpm test` (1920 pass) are all green on this branch while the primitives compile colorless and the app shell is clamped. `environment: 'node'` means Vitest never loads `app.css`, and `e2e/login.spec.ts:16` saves a screenshot but never compares it to a baseline.

**Fix** -- a compile assertion is the cheapest guard, and it turns #2 and #5 red:
```ts
import { compile } from 'tailwindcss';
// compile(readFileSync('src/app.css','utf8')) then .build([...])
// assert bg-primary / text-muted-foreground / bg-popover / border-input / ring-ring / bg-border
//   each emit a non-empty rule
// assert dark:bg-primary emits a `[data-theme` selector, not `prefers-color-scheme`
```
Precedent for a structural guard test: `src/lib/squad/squad-catalog-guard.test.ts`.

src/lib/components/ui/tooltip/tooltip.svelte
```diff
@@ -0,0 +1,7 @@
+<script lang="ts" generics="T = never">
+	import { Tooltip as TooltipPrimitive } from "bits-ui";
+
+	let { open = $bindable(false), ...restProps }: TooltipPrimitive.RootProps<T> = $props();
+</script>
+
+<TooltipPrimitive.Root bind:open {...restProps} />
```
logicminds, line 7: **P2 (#9) -- `Tooltip.Root` throws at init unless a `Tooltip.Provider` is mounted.**

`bits-ui/dist/bits/tooltip/tooltip.svelte.js:175` calls `TooltipProviderContext.get()` eagerly inside `TooltipRootState.create`, and runed's `Context.get()` throws `Context "Tooltip.Provider" not found` when it is absent (`runed/dist/utilities/Context/context.js:40`).

The vendored set exports a Provider but nothing mounts it, and `src/routes/+layout.svelte` has no wrapper. The first `<Tooltip.Root>` a contributor writes is an uncaught init error that tears down the surrounding subtree -- not a missing tooltip.

**Fix** -- one wrapper in `src/routes/+layout.svelte` makes every future `Tooltip.Root` safe with no per-callsite boilerplate:
```svelte
<script lang="ts">
  import { Tooltip } from '$lib/components/ui/tooltip/index.js';
</script>

<Tooltip.Provider>
  <!-- existing layout content -->
</Tooltip.Provider>
```

src/lib/components/ui/dialog/dialog-content.svelte
```diff
@@ -0,0 +1,49 @@
...
+	} = $props();
+</script>
+
+<DialogPortal {...portalProps}>
+	<Dialog.Overlay />
+	<DialogPrimitive.Content
+		bind:ref
+		data-slot="dialog-content"
+		class={cn(
+			"grid max-w-[calc(100%-2rem)] gap-4 rounded-xl bg-popover p-4 text-sm text-popover-foreground ring-1 ring-foreground/10 duration-100 sm:max-w-sm data-open:animate-in data-open:fade-in-0 data-open:zoom-in-95 data-closed:animate-out data-closed:fade-out-0 data-closed:zoom-out-95 fixed top-1/2 left-1/2 z-50 w-full -translate-x-1/2 -translate-y-1/2 outline-none",
```
logicminds: **P2 (#8) -- `z-50` puts portaled primitives below every existing modal, tab and toast.**

bits-ui portals to `document.body`, so these fixed layers land in the root stacking context against our existing overlays: `z-index: 1000` (`src/components/ui/Modal.svelte:110`), `10000` (six rules -- settings export/danger-zone modals, Tab, DefaultWalletConfig), `10001` (two), `99999` (`src/components/ui/Toast.svelte:72`).

The obvious first use -- a DropdownMenu or Popover opened from inside one of our existing modals -- renders underneath it and is unreachable; the user sees the trigger flip to `aria-expanded` with no menu.

**Fix** -- move all five portaled layers onto our scale: `dialog-overlay:15`, `dialog-content:32`, `popover-content:26`, `dropdown-menu-content:26`, `tooltip-content:30`. Suggested `z-[10050]` for dialog overlay/content and `z-[10060]` for menus and tooltips -- above the `10000` modals, below the `99999` toast.

eslint.config.js
```diff
@@ -75,7 +75,10 @@ export default [
 		}
 	},
 	{
-		files: ['src/components/parent/governance/PactoGovInfraList.svelte'],
+		files: [
+			'src/components/parent/governance/PactoGovInfraList.svelte',
+			'src/lib/components/ui/button/button.svelte'
```
logicminds: **P2 (#7) -- permanent exemption parked in the block marked for deletion.**

This block sits under the comment at `eslint.config.js:52-54`: *"Temporary suppressions for files whose remaining lint errors require behavioral fixes. These are being addressed in the parent lint PR and will be removed once the risky changes land."*

The vendored shadcn button needs `svelte/no-navigation-without-resolve` off **permanently** -- `button.svelte:59-63` renders a generic `<a href={disabled ? undefined : href}>` by design, with no SvelteKit route to resolve. Whoever executes that documented cleanup deletes a still-load-bearing override and breaks `pnpm lint` (required before every commit per AGENTS.md) on a file they never touched.

**Fix** -- revert this entry and add a separate permanent block after the temporary section. Directory scoping also covers `badge.svelte:41`'s identical `<svelte:element this={href ? 'a' : 'span'}>` pattern and every primitive vendored later:
```js
// Vendored shadcn-svelte primitives: generic <a href> with no route to resolve.
{
  files: ['src/lib/components/ui/**/*.svelte'],
  rules: { 'svelte/no-navigation-without-resolve': 'off' }
},
```

components.json
```diff
@@ -0,0 +1,16 @@
+{
+  "$schema": "https://shadcn-svelte.com/schema.json",
+  "tailwind": {
+    "css": "src/app.css",
+    "baseColor": "neutral"
+  },
+  "aliases": {
+    "components": "$lib/components",
+    "utils": "$lib/utils",
+    "ui": "$lib/components/ui",
```
logicminds, line 10: **P2 (#6) -- third directory named `ui`, and no rule for which one is canonical.** (Human decision -- not something a reviewer should pick.)

This alias creates `src/lib/components/ui/` alongside the pre-existing `src/components/ui/` (Modal, Toast, Tab, RefreshIconButton, EditIconButton) and `src/lib/ui/`. AGENTS.md is the map contributors are told to start from, and it still says Svelte UI components live under `src/components/` while `src/lib/` holds side-effect modules and typed Tauri wrappers. No doc file is in this diff, so AGENTS.md's own rule -- *"If docs and code disagree, trust the code and update the doc"* -- is unmet.

The next person adding a primitive has three plausible homes and no stated rule.

**Ask** -- decide whether `src/components/ui/` eventually migrates or the two coexist permanently, then record it in AGENTS.md and CLAUDE.md:
```
| `src/lib/components/ui/` | Vendored shadcn-svelte primitives (regenerated via components.json, owned as Pacto code). |
```
and amend the `src/components/` row to scope it to Pacto-authored domain components.

logicminds (changes requested):
## Review summary

Solid direction and the vendoring choice is right -- treating the primitives as Pacto code rather than throwaway CLI output is the correct call. Two things block it, and both trace to the same root: `shadcn-svelte init`'s `app.css` output never came over, and nothing in the toolchain can tell us.

**Verified against this branch's production build** -- `pnpm install --frozen-lockfile`, `pnpm build`, `pnpm check` (2850 files, 0 errors / 0 warnings), `pnpm lint` (clean), `pnpm test` (210 files, 1920 tests, all pass). Everything is green. Every finding below was then confirmed against the shipped stylesheet and by measuring the rendered DOM, because green is exactly the problem here.

### Blocking

| # | Sev | Where | What |
|---|---|---|---|
| 1 | **P0** | `src/app.css:51-58` | Tailwind's `container` utility clamps `<main class="container">`; the shell renders 1280px in a 1365px window, on every screen |
| 2 | **P1** | `src/app.css:1` | No `@theme` block -- all 25 shadcn color utilities compile to nothing, so the primitives render colorless |
| 3 | **P1** | `src/app.css:1` | Preflight drops `<h2>` from 24px to 16px, zeroes block margins, strips list markers -- on screens this PR promises not to change |
| 4 | **P1** | `src/lib/utils.test.ts` | No check anywhere can fail when a Tailwind utility compiles to nothing |

### Non-blocking

| # | Sev | Where | What |
|---|---|---|---|
| 5 | P2 | `avatar.svelte:22` | `dark:` follows the OS, not `[data-theme]`; `techno` is a light theme |
| 6 | P2 | `components.json:10` | Third `ui/` directory, no recorded rule for which is canonical -- needs your call |
| 7 | P2 | `eslint.config.js:80` | Permanent exemption parked in the block documented as temporary |
| 8 | P2 | `dialog-content.svelte:32` | Portaled primitives at `z-50` render under every app modal and toast |
| 9 | P2 | `tooltip.svelte:7` | `Tooltip.Root` throws at init with no `Tooltip.Provider` mounted |

### Suggested order

1. **#1 first** -- `width: 100%; max-width: none;` on `.container` stops a live regression and depends on no other decision.
2. **#2 + #5 together** -- one `@theme inline` + `@custom-variant dark` block in `app.css` closes both.
3. **#4** -- the compile assertion that keeps #2 and #5 from silently regressing again.
4. **#3** -- the one real design call: drop Preflight via the layered import, or keep it and restore the UA behaviours here.
5. **#8 + #9** before any adoption PR; **#7** is a two-minute move; **#6** is yours to decide.

### Worth noting (non-blocking)

- `src/lib/utils.ts` shadows the 13-module `src/lib/utils/` directory. Nothing breaks today, but adding `src/lib/utils/index.ts` would silently re-point ~30 vendored imports.
- `NostrSettingsSection.svelte:783` documents *"this app has no global box-sizing:border-box reset"* as load-bearing rationale. Preflight inverts that; the comment is now false.
- `ui.close` duplicates the existing `commons.close` ("Close"/"Cerrar") in both locales.
- `@internationalized/date` was added to `dependencies` with zero imports in `src/` -- a bits-ui peer for date components none of which were vendored.
- `tailwindcss()` is registered after `sveltekit()` in `vite.config.ts`; Tailwind's SvelteKit guidance puts it first. No breakage observed.
- No `@source` directive, so Tailwind auto-scans from the repo root. ~31KB of the 51.6KB shipped stylesheet is Tailwind output for a layer that currently has zero product consumers.

### Test coverage

No test renders any of the 60 vendored primitives, though `@testing-library/svelte` and `svelteTesting()` are already wired and seven jsdom component tests exist under `src/components/dm/`. Specifically uncovered: `button.svelte:58-66`'s four interlocking disabled-anchor a11y decisions (and `badge.svelte:41`'s repeat of the shape), the `showCloseButton` / `$t("ui.close")` local modification (which is `sr-only`, so invisible to sighted smoke testing and the likeliest casualty of the next shadcn re-sync), and any visual baseline for the Preflight reset.

---

Findings were produced by five independent review passes and each was re-verified by a separate validation pass: 9/9 upheld, 0 dropped. #2 and #5 were each found independently by four of the five passes.

src/app.css
thebeyondr: > Tailwind's `container` utility clamps the app shell on every screen.

Addressed differently in `cfc153d`: renamed the shell class from `container` to `app-shell` in `src/app.css` and `src/routes/+page.svelte`, instead of overriding Tailwind's utility with `max-width: none`. That removes the name collision rather than fighting cascade order.

thebeyondr: > no `@theme` block, so every shadcn color utility compiles to nothing.

Addressed in `cfc153d`: added `@theme inline` in `src/app.css` mapping shadcn semantic colors onto Pacto tokens (`--accent`, `--bg-*`, `--text-*`, etc.). Primary foreground uses a new per-theme `--on-accent` (not `--bg-page`) so light `techno` stays readable.

src/app.css
thebeyondr: > Preflight silently resets headings, block margins and list markers app-wide.

Addressed in `cfc153d` with option 1: import `theme.css` + `utilities.css` only (no global Preflight), matching “no product screens redesigned here.” Vendored primitives get a narrow `[data-slot]` `box-sizing: border-box` reset so they still lay out correctly without rewriting existing screens.

src/lib/components/ui/avatar/avatar.svelte
thebeyondr: > `dark:` keys off the OS color scheme, not our `[data-theme]` switch.

Addressed in `cfc153d`: `@custom-variant dark` is keyed to `dark-techno`, `aztec`, `midnight`, and `union` only. Light `techno` no longer picks up OS dark refinements.

src/lib/utils.test.ts
thebeyondr: > nothing in the toolchain can fail when a Tailwind utility compiles to nothing.

Addressed in `cfc153d`: added `src/lib/ui/tailwind-foundation.test.ts` (layered imports, `@theme` utilities compile, `dark:` emits `[data-theme=` selectors not `prefers-color-scheme`, shell uses `app-shell`). Also added focused primitive tests for disabled-anchor button behavior, dialog close copy, and portal z-index contracts.

src/lib/components/ui/tooltip/tooltip.svelte
thebeyondr: > `Tooltip.Root` throws at init unless a `Tooltip.Provider` is mounted.

Addressed in `cfc153d`: `TooltipProvider` now wraps the full `UpdateGate` tree in `src/routes/+layout.svelte` (auth and logged-in), so the first `Tooltip.Root` use cannot miss provider context.

src/lib/components/ui/dialog/dialog-content.svelte
thebeyondr: > `z-50` puts portaled primitives below every existing modal, tab and toast.

Addressed in `cfc153d`: dialog overlay/content use `z-[10050]`; dropdown, popover, and tooltip content use `z-[10060]` — above existing ~10000 modals, below toast (`99999`).

eslint.config.js
thebeyondr: > permanent exemption parked in the block marked for deletion.

Addressed in `cfc153d`: moved the shadcn override into its own permanent block for `button.svelte` and `badge.svelte`, and left the temporary suppressions block alone.

components.json
thebeyondr: > third directory named `ui`, and no rule for which one is canonical.

Addressed in `cfc153d`: documented coexistence in `AGENTS.md` / `CLAUDE.md` — `src/lib/components/ui/` is vendored shadcn primitives; `src/components/` stays Pacto-authored domain/shell UI. No forced migration in this foundation PR.

thebeyondr:
> Review summary on `build/ui-foundation` (changes requested)

Addressed the blocking and P2 adoption items in `cfc153d` (now on `origin/build/ui-foundation`). Thread-by-thread replies are on the inline findings.

**Worth noting (intentionally not changed here):**
- `src/lib/utils.ts` vs `src/lib/utils/`: no `index.ts` today; Node resolution keeps `$lib/utils` on the file. Adding an `index.ts` later would need care — left as-is.
- `@internationalized/date`: required Bits UI peer dependency even without date primitives vendored yet.
- Duplicate `ui.close` vs `commons.close`: keeping a `ui.*` namespace isolates vendored copy from domain catalogs.
- Vite plugin order / no `@source` / CSS size: no observed breakage; narrowing sources can wait until product screens adopt primitives.
- Stale box-sizing comment in `NostrSettingsSection`: still accurate after dropping global Preflight.

Happy to take a re-review pass.

logicminds (approved):
LGTM

PR #262, opened 2026-08-13
feat: seed a populated sandbox with no docker and no network

logicminds:
## Summary

Getting a populated dev account required the full local stack: docker, a relay, and a live MLS round trip. On a machine without docker there was no option at all, which made the "populated sandbox" story unavailable exactly where it was needed most.

After this change, one command produces a populated per-account database with **zero network and no docker**, built through the real ingest paths rather than hand-written rows.

## Changes

- **Harness binary** linking the library crate, with no Tauri IPC and no window; a mock app handle serves purely as a path resolver.
- **DM slice** drives `rumor::process_rumor` directly and persists via the normal `StoredEvent` path.
- **Squad slice** stands up a second in-process MDK engine that plays the inviter, produces a real welcome, and feeds it through normal inbound processing. No committed wire-byte fixtures - they are both a crypto hazard and a drift source.
- **Wallet state** seeded through the existing BIP-44 derivation, not direct table writes.
- **Attachments** skipped with a recorded reason, since that path needs the process-global Tauri handle a mock handle can never populate.
- Gated behind a non-default `relay-free-harness` feature with `required-features`, plus a CI job that runs it with outbound networking dropped.

## Manifest version handshake (out of scope for this unit)

The plan's verification contract lists a manifest version-handshake row against U6, U7, and U14 (this unit): load a manifest one version above and one below each consumer's window and confirm every consumer refuses loudly rather than silently degrading. This harness isn't a manifest consumer at all - its CLI is flat scalars, `relay-free-harness [--sandbox-root PATH] [--mnemonic PHRASE] [--pin PIN]` (`src-tauri/src/bin/relay_free_harness.rs:9`), and no manifest file is parsed anywhere; the orchestrator passes the phrase *from* a manifest, but the harness itself never sees one. So that contract row is satisfied vacuously here, not because the handshake was implemented and exercised. If the harness ever grows manifest input, the version-handshake obligation comes with it.

## Test plan

- `cargo build --no-default-features --features relay-free-harness --bin relay-free-harness`
- `cargo test --lib --no-default-features --features relay-free-harness harness::` - 5/5
- `cargo test --lib --no-default-features` - 663 passed, 0 failed (no regression)
- Ran the binary against a fresh sandbox root, then inspected the SQLite directly: **6 events, 2 chats, 1 mls_groups row, 1 evm_accounts row**, `evm_address` matching the canonical BIP-44 account for the test mnemonic.
- Reran against the same root: idempotent, counts unchanged. Corrupting the completion marker produces a clear refusal rather than a silent reseed.
- Deleted the built binary and ran `cargo build --bins` with default features: the harness is **absent**, confirming exclusion from default and release builds.
- Refuses immediately when no sandbox root is set, so it can never fall through to the real OS data directory.

---

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -221,20 +221,27 @@ jobs:
 
       - name: Seed a sandbox with no network reachable
         working-directory: src-tauri
+        # The CI image's default shell is dash, which has no `pipefail`, so
+        # redirect rather than pipe -- `set -e` then still fails the step on a
+        # non-zero harness exit instead of only on a failed `grep`.
         run: |
-          set -euo pipefail
+          set -eu
           mkdir -p "$RUNNER_TEMP/relay-free-harness-sandbox"
           iptables -A OUTPUT -j DROP
-          ./target/debug/relay-free-harness --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
-            | tee /tmp/harness-first-run.log
+          ./target/debug/relay-free-harness \
+            --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+            > /tmp/harness-first-run.log 2>&1
+          cat /tmp/harness-first-run.log
           grep -q "Seeded npub=" /tmp/harness-first-run.log
 
       - name: Rerun is idempotent, still with no network reachable
         working-directory: src-tauri
         run: |
-          set -euo pipefail
-          ./target/debug/relay-free-harness --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
-            | tee /tmp/harness-second-run.log
+          set -eu
+          ./target/debug/relay-free-harness \
+            --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+            > /tmp/harness-second-run.log 2>&1
+          cat /tmp/harness-second-run.log
           grep -q "already seeded" /tmp/harness-second-run.log
 
   landing-build:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -195,6 +195,48 @@ jobs:
         working-directory: src-tauri
         run: cargo test --lib --no-default-features
 
+  relay-free-harness:
+    # Relay-free seeding harness: builds populated per-account storage with
+    # zero network calls, through the real ingest path.
+    # No dev-env docker-compose stack. `iptables -j DROP` on OUTPUT proves the
+    # binary's own "zero network calls" claim at runtime, not just an
+    # assertion -- NET_ADMIN is the one added capability it needs.
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/covenant-gov/pacto-app/pacto-ci:latest
+      options: --user root --cap-add=NET_ADMIN
+    steps:
+      - uses: actions/checkout@v6
+
+      - uses: Swatinem/rust-cache@v2
+        with:
+          workspaces: src-tauri
+
+      - name: Build relay-free harness
+        working-directory: src-tauri
+        run: cargo build --no-default-features --features relay-free-harness --bin relay-free-harness
+
+      - name: Install iptables
+        run: apt-get update -qq && apt-get install -y -qq iptables
+
+      - name: Seed a sandbox with no network reachable
+        working-directory: src-tauri
+        run: |
+          set -euo pipefail
+          mkdir -p "$RUNNER_TEMP/relay-free-harness-sandbox"
+          iptables -A OUTPUT -j DROP
+          ./target/debug/relay-free-harness --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+            | tee /tmp/harness-first-run.log
+          grep -q "Seeded npub=" /tmp/harness-first-run.log
+
+      - name: Rerun is idempotent, still with no network reachable
+        working-directory: src-tauri
+        run: |
+          set -euo pipefail
+          ./target/debug/relay-free-harness --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+            | tee /tmp/harness-second-run.log
+          grep -q "already seeded" /tmp/harness-second-run.log
+
   landing-build:
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -198,13 +198,14 @@ jobs:
   relay-free-harness:
     # Relay-free seeding harness: builds populated per-account storage with
     # zero network calls, through the real ingest path.
-    # No dev-env docker-compose stack. `iptables -j DROP` on OUTPUT proves the
-    # binary's own "zero network calls" claim at runtime, not just an
-    # assertion -- NET_ADMIN is the one added capability it needs.
+    # No dev-env docker-compose stack. The harness binary alone runs inside
+    # `unshare --net` so checkout / rust-cache / apt keep network, while the
+    # seed itself has an empty network namespace. This is an honest-binary
+    # smoke test (the PR-built binary is not adversarially sandboxed).
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/covenant-gov/pacto-app/pacto-ci:latest
-      options: --user root --cap-add=NET_ADMIN
+      options: --user root --cap-add=SYS_ADMIN
     steps:
       - uses: actions/checkout@v6
 
@@ -212,34 +213,49 @@ jobs:
         with:
           workspaces: src-tauri
 
+      - name: Harness unit tests
+        working-directory: src-tauri
+        run: cargo test --lib --no-default-features --features relay-free-harness harness::
+
       - name: Build relay-free harness
         working-directory: src-tauri
         run: cargo build --no-default-features --features relay-free-harness --bin relay-free-harness
 
-      - name: Install iptables
-        run: apt-get update -qq && apt-get install -y -qq iptables
+      - name: Install sqlite3
+        run: apt-get update -qq && apt-get install -y -qq sqlite3
 
-      - name: Seed a sandbox with no network reachable
+      - name: Seed a sandbox with no network namespace
         working-directory: src-tauri
         # The CI image's default shell is dash, which has no `pipefail`, so
         # redirect rather than pipe -- `set -e` then still fails the step on a
         # non-zero harness exit instead of only on a failed `grep`.
+        # Root must sit under test_sandbox/ (harness placement check).
         run: |
           set -eu
-          mkdir -p "$RUNNER_TEMP/relay-free-harness-sandbox"
-          iptables -A OUTPUT -j DROP
-          ./target/debug/relay-free-harness \
-            --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+          SANDBOX_ROOT="$GITHUB_WORKSPACE/test_sandbox/relay-free-harness-ci"
+          mkdir -p "$SANDBOX_ROOT"
+          unshare --net -- ./target/debug/relay-free-harness \
+            --sandbox-root "$SANDBOX_ROOT" \
             > /tmp/harness-first-run.log 2>&1
           cat /tmp/harness-first-run.log
           grep -q "Seeded npub=" /tmp/harness-first-run.log
+          grep -q "dm_messages=2" /tmp/harness-first-run.log
+          grep -E -q 'squad_group_id=Some\("[^"]+"\)' /tmp/harness-first-run.log
+          # sqlite3-assert the counts already documented in the PR body.
+          ACCOUNT_DB=$(find "$SANDBOX_ROOT" -name 'pacto.db' | head -n 1)
+          test -n "$ACCOUNT_DB"
+          test "$(sqlite3 "$ACCOUNT_DB" 'SELECT COUNT(*) FROM events;')" -eq 6
+          test "$(sqlite3 "$ACCOUNT_DB" 'SELECT COUNT(*) FROM chats;')" -eq 2
+          test "$(sqlite3 "$ACCOUNT_DB" 'SELECT COUNT(*) FROM mls_groups;')" -eq 1
+          test "$(sqlite3 "$ACCOUNT_DB" 'SELECT COUNT(*) FROM evm_accounts;')" -eq 1
 
-      - name: Rerun is idempotent, still with no network reachable
+      - name: Rerun is idempotent, still with no network namespace
         working-directory: src-tauri
         run: |
           set -eu
-          ./target/debug/relay-free-harness \
-            --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+          SANDBOX_ROOT="$GITHUB_WORKSPACE/test_sandbox/relay-free-harness-ci"
+          unshare --net -- ./target/debug/relay-free-harness \
+            --sandbox-root "$SANDBOX_ROOT" \
             > /tmp/harness-second-run.log 2>&1
           cat /tmp/harness-second-run.log
           grep -q "already seeded" /tmp/harness-second-run.log
--- a/src-tauri/src/bin/relay_free_harness.rs
+++ b/src-tauri/src/bin/relay_free_harness.rs
@@ -1,0 +1,102 @@
+//! Thin entry point for the relay-free seeding harness.
+//!
+//! No Tauri IPC and no window: a `tauri::test` mock app handle serves purely
+//! as a path resolver, which the sandbox-root override in
+//! `test_sandbox::sandbox_root` already bypasses whenever
+//! `PACTO_TEST_SANDBOX_ROOT` is set. All real work lives in
+//! `pacto_lib::harness`; see that module for the seeding logic itself.
+//!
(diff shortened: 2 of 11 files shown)


src-tauri/src/harness.rs
```diff
@@ -0,0 +1,618 @@
...
+/// the harness's primary idempotency guard.
+const SEED_MARKER_KEY: &str = "harness_seed_complete";
+const SEED_MARKER_VERSION: &str = "1";
+
+/// Anvil/Hardhat's well-known public test mnemonic -- already the default
+/// local chain this repo's sandboxes point at. A fine default: this identity
+/// is a seeding fixture, not a real account, so its key material is public
+/// by construction. Override with `--mnemonic`.
+pub const DEFAULT_MNEMONIC: &str = "test test test test test test test test test test test junk";
+pub const DEFAULT_PIN: &str = "123456";
```
daopunk, line 53: **P1 — public fixture identity, no sandbox-only stamp.**

This is the well-known Anvil/Hardhat phrase (same BIP-44 account 0 already documented as public). KD9 / R25 in `docs/plans/2026-08-10-001-feat-parallel-agent-sandboxes-plan.md` require a recipe-derived identity to be stamped sandbox-only, and the app must refuse it while any non-local relay is in the set. `dev_login` already enforces that via `PACTO_DEV_IDENTITY_SANDBOX_ONLY`.

The harness never stamps that mark. A developer who seeds, then boots `make dev-sandbox` / `dev_login` on this root against the compiled production relay set, will use a publicly-derivable key as a real account. That is exactly the case KD9 forbids.

Also: `--mnemonic` is documented as a supported override with default PIN `123456`. If someone points this at a non-fixture phrase, the seed is persisted under a trivial PIN.

**Ask:** Stamp the seeded identity with the same sandbox-only mark `dev_login` honors. Refuse a non-fixture `--mnemonic` unless a second explicit flag is passed. Never print the mnemonic. Document that opening this DB in the live app still requires `PACTO_TRUSTED_RELAYS` (local) + `PACTO_DEV_IDENTITY_SANDBOX_ONLY`.

src-tauri/src/harness.rs
```diff
@@ -0,0 +1,618 @@
...
+    let inviter_keys = derive_synthetic_keys("squad-inviter/v1");
+    let relays: Vec<RelayUrl> = crate::trusted_relays::trusted_relays().to_vec();
+    let group_config = NostrGroupConfigData::new(
+        SQUAD_NAME.to_string(),
+        "Seeded by the relay-free harness".to_string(),
+        None,
+        None,
+        None,
+        relays,
+        vec![inviter_keys.public_key()],
```
daopunk, line 499: **P1 — production relays embedded in a relay-free seed.**

`trusted_relays::trusted_relays()` here is the compiled default set (`jskitty.cat`, `asia.vectorapp.io`, `computingcache.com`). The harness never calls `trusted_relays::init_from_env()`, so a local override is ignored even if the operator set one.

Those URLs land in MLS group config metadata inside the persistent store. Opening this DB later with network up will point the group at production relays, using the public Anvil-derived identity from the comment above.

For a *relay-free* seed, use an empty list, a documented local placeholder, or whatever `all_relays_local()` would accept — not the production set.

`create_key_package_for_event(..., trusted_relays())` a few dozen lines below has the same problem.

src-tauri/src/harness.rs
```diff
@@ -0,0 +1,618 @@
...
+/// (not just by a Makefile/CI wrapper) so running the binary directly can
+/// never fall through to the real OS app-data directory -- the same
+/// discipline `test_sandbox::enforce_dev_world_root` applies to the main app.
+pub fn require_sandbox_root() -> Result<std::path::PathBuf, String> {
+    crate::test_sandbox::sandbox_root().ok_or_else(|| {
+        "PACTO_TEST_SANDBOX_ROOT is not set; the relay-free harness refuses to run without an \
+         explicit sandbox root rather than risk touching the real OS data directory."
+            .to_string()
+    })
+}
```
daopunk: **P1 — presence-only check, not a sandbox check.**

This fails when the env var is unset (good), but any non-blank path is accepted. `--sandbox-root` / `PACTO_TEST_SANDBOX_ROOT` can be the real OS app-data dir, a `make dev` tree, or `$HOME/Library/Application Support/…`.

`test_sandbox::enforce_dev_world_root` actually validates placement. This does not, despite the comment claiming the same discipline.

Combined with default PIN `123456` and optional `--mnemonic`, this is how a real profile DB gets overwritten.

**Ask:** Reject roots that are not under `test_sandbox/` / `test_fixtures/` (or reuse the `enforce_dev_world_root` placement check).

src-tauri/src/bin/relay_free_harness.rs
```diff
@@ -0,0 +1,102 @@
...
+                sandbox_root = Some(args.next().unwrap_or_else(|| {
+                    eprintln!("--sandbox-root requires a path argument");
+                    std::process::exit(2);
+                }));
+            }
+            "--mnemonic" => {
+                config.mnemonic = args.next().unwrap_or_else(|| {
+                    eprintln!("--mnemonic requires a phrase argument");
+                    std::process::exit(2);
+                });
```
daopunk: **P1 — mnemonic on argv.**

`--mnemonic` puts a recovery phrase in the process list (`ps`, CI logs if anyone echoes `$@`, shell history). The default is a public fixture so CI is fine; the flag exists so operators can pass a *real* phrase.

Refuse a non-fixture phrase unless a second explicit flag is passed. Prefer `PACTO_DEV_LOGIN_MNEMONIC` (already the repo convention) over argv. Never print the phrase.

.github/workflows/ci.yaml
```diff
@@ -195,6 +195,55 @@ jobs:
...
+  relay-free-harness:
+    # Relay-free seeding harness: builds populated per-account storage with
+    # zero network calls, through the real ingest path.
+    # No dev-env docker-compose stack. `iptables -j DROP` on OUTPUT proves the
+    # binary's own "zero network calls" claim at runtime, not just an
+    # assertion -- NET_ADMIN is the one added capability it needs.
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/covenant-gov/pacto-app/pacto-ci:latest
+      options: --user root --cap-add=NET_ADMIN
```
daopunk: **P1 — `OUTPUT DROP` is not scoped, and `NET_ADMIN`+root is not a sandbox.**

- `iptables -A OUTPUT -j DROP` is never deleted. `Swatinem/rust-cache@v2` saves cache in a post-job hook in the same network namespace; DROP can fail that upload (flake, or silent cache miss forever).
- `iptables` is IPv4-only. IPv6 OUTPUT can still leave the box, so the "zero network" proof is incomplete.
- The binary is compiled from PR code, then run as **root with `NET_ADMIN`**. It can `iptables -F` and restore the network. This job proves the *honest* binary does not need the network. It is not an adversarial sandbox for untrusted PRs — please do not imply that.

**Ask:** Prefer `unshare --net` (or equivalent) around *only* the harness binary so checkout/cache/apt are unaffected and `NET_ADMIN` is unnecessary. If iptables stays: also drop IPv6 (`ip6tables` / inet table), and delete the rules at the end of the last harness step.

.github/workflows/ci.yaml
```diff
@@ -195,6 +195,55 @@ jobs:
...
+
+      - name: Rerun is idempotent, still with no network reachable
+        working-directory: src-tauri
+        run: |
+          set -eu
+          ./target/debug/relay-free-harness \
+            --sandbox-root "$RUNNER_TEMP/relay-free-harness-sandbox" \
+            > /tmp/harness-second-run.log 2>&1
+          cat /tmp/harness-second-run.log
+          grep -q "already seeded" /tmp/harness-second-run.log
```
daopunk: **P2 — CI never runs `harness::` tests and barely asserts seeding.**

The PR test plan includes `cargo test --lib --no-default-features --features relay-free-harness harness::` (5/5). This job does not run it. `backend-tests` still uses `--no-default-features` without the feature, so those tests never run in CI.

The job greps `Seeded npub=` / `already seeded`. It does not check `dm_messages=2`, a non-empty `squad_group_id`, or the SQLite counts already in the PR body (6 events, 2 chats, 1 `mls_groups` row, 1 `evm_accounts` row).

`seed_squad_slice` silently ignores non-`ApplicationMessage` results, so a future MDK change can "succeed" with an empty squad chat and still pass this grep.

**Ask:** Run the unit tests in this job (or `backend-tests` with the feature). After seed, sqlite3-assert those counts. Fail the step if the first-run log does not contain `dm_messages=2` and a real `squad_group_id`.

src-tauri/src/harness.rs
```diff
@@ -0,0 +1,618 @@
...
+) -> Result<Option<String>, String> {
+    if let Some(existing) = db::load_mls_groups(handle)
+        .await?
+        .into_iter()
+        .find(|g| g.name == SQUAD_NAME)
+    {
+        return Ok(Some(existing.group_id));
+    }
+
+    let sandbox_kp_event = generate_and_store_device_keypackage(handle, keys).await?;
```
daopunk, line 486: **P2 — crash mid-squad is not idempotent, despite this comment.**

The SQL-name guard and the top-level `harness_seed_complete` marker (written last) do not cover: process dies after `accept_welcome` (persistent MLS store mutated) but before `db::save_mls_group`.

A rerun then:

1. Does not see the `mls_groups` row.
2. Calls `create_key_package_for_event` again (`save_mls_keypackages` DELETEs the index first).
3. Inviter `create_group` with a new random MLS group id — the thing this comment says must never happen twice.
4. `process_welcome` on an engine that already has a group.

CI will not hit this (fresh root). Local / agent reruns will.

**Ask:** Treat "MLS store already has a group" as seeded, or wipe the MLS db on incomplete seed. Do not mint a second keypackage unless the store is empty. Also fail rather than `unwrap_or_else(|| nostr_group_id.clone())` if `get_groups()` misses the group (~L461), and error if `process_message` does not yield `ApplicationMessage` (~L489).

daopunk (changes requested):
## Verdict: request changes

The ingest-path approach is the right one: real `rumor::process_rumor`, a second in-process MDK engine for the welcome, `db::*` helpers, no committed wire-byte fixtures, and a non-default `required-features` gate so default/release builds skip the binary. Unset-root refusal and unexpected marker values are the right instincts. MDK `!Send` is handled. CI builds before the network drop, and the dash/`pipefail` follow-up shows the job was actually run.

Do not merge until the public-identity/relay issue is addressed, and until the CI network-drop is scoped so it cannot break cache upload or miss IPv6.

## What is good

- Seeds through the real ingest path rather than hand-written rows.
- Feature + `required-features` keeps the binary out of `cargo build --bins` and Tauri release.
- Marker mismatch refuses rather than silently reseeding.
- Engine scopes drop before await.

## P1

1. **Public fixture identity vs production relays.** Default mnemonic is the well-known Anvil phrase. Squad `NostrGroupConfigData` is built with compiled `trusted_relays()` (`jskitty.cat`, `asia.vectorapp.io`, `computingcache.com`). No `init_from_env()`, no sandbox-only stamp, no refusal if this DB is later opened by `make dev-sandbox` / `dev_login` against the production relay set. That is the KD9 / R25 case `dev_login` already blocks via `PACTO_DEV_IDENTITY_SANDBOX_ONLY`.

2. **`iptables -A OUTPUT -j DROP` is process-wide and never undone.** It survives into the rust-cache post-job hook, is IPv4-only, and the PR-built binary then runs as root with `NET_ADMIN` (so it can `iptables -F`). This is an honest-binary smoke test, not an adversarial sandbox.

3. **Sandbox root is presence-only.** Any `--sandbox-root` / `PACTO_TEST_SANDBOX_ROOT` is accepted, including a real OS data dir. Combined with `--mnemonic` and default PIN `123456`, that can encrypt a real seed into a real profile tree.

## P2 (fix or explicitly defer)

4. Crash after `accept_welcome` but before `save_mls_group` is not idempotent: rerun mints a new keypackage and calls `create_group` against an engine that already has a group.
5. CI never runs `harness::` unit tests. The job greps log strings and does not sqlite3-assert the counts already in the PR body. Non-`ApplicationMessage` MLS results are silently skipped; `engine_group_id` falls back to `nostr_group_id`.
6. No Makefile / `DEV_SANDBOX.md` / `AGENTS.md` entry point. `SEED_MARKER_VERSION` will rot across contributors. `check-release-symbols.sh` does not look for harness strings, so "never in a release build" is policy, not a gate.
7. Attachments skipped (honest). Squad `Chat` is created with empty participants, so the UI can look memberless.

## P3 — merge / upgrade / multi-dev

File overlap vs `main` is clean (0 behind). `lib.rs` hunks vs #264 / #243 are disjoint/additive. The real cost is **behavioral coupling**: later `rumor` / `MlsService` / `db` / MDK 0.8.0 / migration changes can break this CI job without touching these five files. Treat the harness as an ingest+MLS+migrations canary and say so in docs.

MDK 0.9 (`cgka-engine`) will require a rewrite of the second-engine slice. This bin is also the first non-`#[cfg(test)]` consumer of `tauri::test::mock_builder`. Edition 2024 makes `std::env::set_var` unsafe.

Every default run is the same npub. Fine for fixtures; disastrous if that npub ever hits a shared relay (back to P1).

Inline comments cover P1 and the P2 items that belong on specific hunks.

src-tauri/src/harness.rs
logicminds: > Stamp the seeded identity with the same sandbox-only mark `dev_login` honors. Refuse a non-fixture `--mnemonic` unless a second explicit flag is passed. Never print the mnemonic.

Addressed: the harness now writes `.pacto_dev_identity_sandbox_only` under the sandbox root plus a `dev_identity_sandbox_only` SQL setting; `dev_login` treats the marker like `PACTO_DEV_IDENTITY_SANDBOX_ONLY=1`. Non-fixture mnemonics require `--allow-non-fixture-mnemonic` / `PACTO_HARNESS_ALLOW_NON_FIXTURE_MNEMONIC=1`. Opening docs note local `PACTO_TRUSTED_RELAYS` + the sandbox-only stamp/env.

src-tauri/src/harness.rs
logicminds: > For a *relay-free* seed, use an empty list, a documented local placeholder, or whatever `all_relays_local()` would accept — not the production set.

Addressed: MLS group config and `create_key_package_for_event` now use `ws://127.0.0.1:1` (`harness_local_relays()`), which `all_relays_local()` accepts. The compiled production set is no longer embedded.

src-tauri/src/harness.rs
logicminds: > Reject roots that are not under `test_sandbox/` / `test_fixtures/` (or reuse the `enforce_dev_world_root` placement check).

Addressed: `require_sandbox_root()` now requires a `test_sandbox` or `test_fixtures` path component, so a real OS app-data dir is refused.

src-tauri/src/bin/relay_free_harness.rs
logicminds: > Prefer `PACTO_DEV_LOGIN_MNEMONIC` (already the repo convention) over argv. Never print the phrase.

Addressed: mnemonic resolution prefers `PACTO_DEV_LOGIN_MNEMONIC`; non-fixture phrases need `--allow-non-fixture-mnemonic`. The phrase is never printed.

.github/workflows/ci.yaml
logicminds: > Prefer `unshare --net` (or equivalent) around *only* the harness binary so checkout/cache/apt are unaffected and `NET_ADMIN` is unnecessary.

Addressed: dropped iptables/`NET_ADMIN`. The harness binary alone runs under `unshare --net` (job needs `SYS_ADMIN` for that). Comments now describe this as an honest-binary smoke test, not an adversarial sandbox.

.github/workflows/ci.yaml
logicminds: > Run the unit tests in this job (or `backend-tests` with the feature). After seed, sqlite3-assert those counts. Fail the step if the first-run log does not contain `dm_messages=2` and a real `squad_group_id`.

Addressed: the job runs `cargo test … harness::`, greps `dm_messages=2` and a non-empty `squad_group_id`, and sqlite3-asserts events=6 / chats=2 / mls_groups=1 / evm_accounts=1. Non-`ApplicationMessage` MLS results now fail the seed instead of being skipped.

src-tauri/src/harness.rs
logicminds: > Treat "MLS store already has a group" as seeded, or wipe the MLS db on incomplete seed. Do not mint a second keypackage unless the store is empty.

Addressed: `recover_squad_from_mls_store` finishes the SQL side when the persistent MLS store already has a group, and skips minting another keypackage / `create_group`. Missing `get_groups()` match and non-`ApplicationMessage` results are hard errors.

logicminds:
> 6. No Makefile / `DEV_SANDBOX.md` / `AGENTS.md` entry point. `SEED_MARKER_VERSION` will rot across contributors. `check-release-symbols.sh` does not look for harness strings…

Addressed (P2 #6): added `make relay-free-harness-seed`, documented the entry point + marker/canary notes in `docs/build/DEV_SANDBOX.md` and `AGENTS.md`, and extended `scripts/check-release-symbols.sh` with harness strings. P2 #7 (empty squad participants) is also fixed in the seed path.

daopunk (approved):

PR #263, opened 2026-08-13
feat: quarantine stale sandbox profiles instead of bricking boot

logicminds:
## Summary

Hopping between branches whose schemas differ leaves a sandbox profile the current build cannot read. Recovery was a five-step manual `sqlite3` runbook - fine for a human who knows it exists, unusable by an unattended agent, which simply stalls on the update-required screen forever.

- **Before:** a branch hop produces an unreadable sandbox profile and boot stops until someone runs the manual runbook.
- **After:** the offending profile is moved aside automatically, boot proceeds against a fresh profile, and a record explains what was moved and why.

## Changes

- The existing pre-auth scan now quarantines before building the compatibility report, **reusing the verdicts the scan already produces** rather than introducing a second notion of staleness.
- Quarantine is scoped to an active sandbox root. Against the real OS data directory nothing moves and the existing update-required screen is unchanged - **a developer's real account can never be touched.**
- Immediately before the rename, the path is re-checked as a real directory inside the sandbox root and not a symlink, closing the window between classification and move.
- Destination names carry a timestamp plus an atomic sequence, so quarantining twice in one boot cannot collide.
- The manual runbook is retired from `docs/build/OPERATOR_UPDATES.md`.

No new Tauri command: the doctor runs inside the existing `get_storage_compatibility` implementation, which the startup update gate already calls, so there is no orphaned-command surface to wire up.

## Test plan

- `cargo test --lib storage_format` - **32 passed**, including the pre-existing `embedded_set_matches_committed_migration_files` backstop.
- One test per acceptance case: over-ceiling history quarantined; divergent checksum quarantined; healthy profile untouched; only the stale one moves when a healthy profile sits beside it; **no sandbox root means report-only, nothing moved**; a symlink swapped in between classification and move is refused; the record names profile, verdict and offending version; two quarantines in one boot do not collide.
- `cargo test --lib` - 663 passed, 0 failed (no regression).
- `pnpm check` 0 errors; `pnpm lint` matches `main`'s baseline; clippy unchanged at baseline.

---

### DIFF
--- a/docs/build/OPERATOR_UPDATES.md
+++ b/docs/build/OPERATOR_UPDATES.md
@@ -151,26 +151,15 @@ Use these in priority order.
 
    **After using either path to correct a value, mirror the final value back into the tracked `scripts/release-compatibility.json` and commit it.** The workflow and the local script both operate on the *published* `latest.json` only; they never write the tracked file. If the tracked file is left at the old value, the next tagged release re-runs the automatic stamping job with that stale value and silently regresses the correction.
 
-## Recovering a machine the local storage-format check blocked
+## Automatic recovery from a stale sandbox profile
 
-This is a **different** trigger from the minimum-version check above: it never touches the network. It fires when the app opens a profile's `pacto.db` on launch and finds a migration history that this build doesn't recognize — for example, a profile last opened by a newer or divergent build. Because the check runs before account selection, one bad profile blocks **every** account on that machine, not just the one it belongs to.
+This is a **different** trigger from the minimum-version check above: it never touches the network. It fires when the app opens a profile's `pacto.db` on launch and finds a migration history this build doesn't recognize -- for example, a profile last opened by a newer or divergent build. Because the check runs before account selection, one bad profile blocks **every** account on that machine, not just the one it belongs to.
 
-The block screen deliberately does not show which npub or file path is at fault, so recovery is manual:
+**Inside any sandbox** (`make dev` on a branch other than `main`, `make dev-sandbox`, `make dev-world`, or anywhere else `PACTO_TEST_SANDBOX_ROOT` is set), this is handled automatically. The storage doctor moves the offending profile directory aside under `<sandbox_root>/quarantine/` before the compatibility check runs, and boot proceeds against a fresh profile -- no manual step, and no other account on that sandbox is blocked. What moved and why is recorded in `<sandbox_root>/quarantine-record.json` (the profile, the verdict, and the offending schema version), appended to on every quarantine so the full history survives across boots.
 
-1. Note the schema version number the block screen reports.
-2. Profile directories live under `<app_data_dir>/npub1…/pacto.db` — see [`../storage-layout/SQLITE_AND_FILES.md`](../storage-layout/SQLITE_AND_FILES.md) for the full on-disk layout. `<app_data_dir>` here is Tauri's per-OS app data directory *including* this app's identifier (`io.pacto`, from `src-tauri/tauri.conf.json`):
-   - macOS: `~/Library/Application Support/io.pacto/`
-   - Windows: `%APPDATA%\io.pacto\`
-   - Linux: `$XDG_DATA_HOME/io.pacto/` (usually `~/.local/share/io.pacto/`)
-3. For each `npub1…` subdirectory, open its `pacto.db` with a `sqlite3` client and check the highest applied version:
-   ```bash
-   sqlite3 "<app_data_dir>/npub1exampleaddress/pacto.db" \
-     "SELECT version, name, applied_on FROM refinery_schema_history ORDER BY version DESC LIMIT 1;"
-   ```
-4. The one profile whose `version` matches the number from the block screen is the offender. Move that single directory aside (e.g. rename `npub1…` to `npub1…-quarantined`) — do not delete it.
-5. Relaunch the app. Every other account on the machine unblocks immediately; the moved-aside profile no longer participates in account discovery until it is restored (e.g. by a build that recognizes its schema).
+**On `main` (the real OS data directory), quarantine never runs.** This is by construction: the doctor only acts inside a sandbox root, so it can never touch the real account. Hitting this block on `main` means the installed build is older than the schema its own account database was last written with (most commonly after downgrading, or running a newer build against it once) -- the fix is to install a build that recognizes that schema. Profile directories live under `<app_data_dir>/npub1…/pacto.db` -- see [`../storage-layout/SQLITE_AND_FILES.md`](../storage-layout/SQLITE_AND_FILES.md) for the on-disk layout and each platform's `<app_data_dir>`.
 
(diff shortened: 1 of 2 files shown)


docs/build/OPERATOR_UPDATES.md
```diff
@@ -151,26 +151,15 @@ Use these in priority order.
...
-     "SELECT version, name, applied_on FROM refinery_schema_history ORDER BY version DESC LIMIT 1;"
-   ```
-4. The one profile whose `version` matches the number from the block screen is the offender. Move that single directory aside (e.g. rename `npub1…` to `npub1…-quarantined`) — do not delete it.
-5. Relaunch the app. Every other account on the machine unblocks immediately; the moved-aside profile no longer participates in account discovery until it is restored (e.g. by a build that recognizes its schema).
+**On `main` (the real OS data directory), quarantine never runs.** This is by construction: the doctor only acts inside a sandbox root, so it can never touch the real account. Hitting this block on `main` means the installed build is older than the schema its own account database was last written with (most commonly after downgrading, or running a newer build against it once) -- the fix is to install a build that recognizes that schema. Profile directories live under `<app_data_dir>/npub1…/pacto.db` -- see [`../storage-layout/SQLITE_AND_FILES.md`](../storage-layout/SQLITE_AND_FILES.md) for the on-disk layout and each platform's `<app_data_dir>`.
```
daopunk: The five-step sqlite3 runbook is still the only recovery path for the real OS data directory. The block screen carries no npub or path, and the doctor does not run when `PACTO_TEST_SANDBOX_ROOT` is unset, so a machine with one stale profile among several is stuck with no documented way to identify which `npub1…` directory to move aside.

Please restore the runbook under a heading for the real OS data directory (keep the new sandbox auto-recovery as its own subsection). Also: the gate is the env var, not git branch `main`. `make dev` on `main` usually leaves it unset, but that is Makefile convention — say `PACTO_TEST_SANDBOX_ROOT` is the contract.

src-tauri/src/storage_format.rs
```diff
@@ -233,6 +242,215 @@ pub(crate) fn scan_profiles(app_data_dir: &Path) -> StorageFormatScanReport {
...
+/// existing compatibility report to name, exactly today's behavior, and
+/// quarantine can therefore never touch the real OS-data account.
+///
+/// A profile is offending exactly when `classify_database` -- the same
+/// per-profile classification `scan_profiles` already runs -- returns
+/// `Unrecognized` or `Divergent`; every other verdict is left alone.
+fn quarantine_stale_profiles(app_data_dir: &Path) {
+    let Some(sandbox_root) = crate::test_sandbox::sandbox_root() else {
+        return;
+    };
```
daopunk, line 465: The early return is “env var is set”, then this walks whatever `app_data_dir` was passed. Per-profile `revalidate_before_move` is good, but `compatibility_report` is now destructive and any caller (tests, a future command) can pass an arbitrary path while the env is set.

Refuse the whole scan unless `app_data_dir` itself canonicalizes under the sandbox root (ideally `<root>/data`) before classifying anything. Add a test: env set, scanned directory *outside* that root, nothing moved. The symlink-swap test covers a different case.

`cfg!(debug_assertions) && sandbox_root().is_some()` matching `multi_instance_allowed` would be belt-and-suspenders for R18; not required for correctness on the production path.

src-tauri/src/storage_format.rs
```diff
@@ -265,6 +483,7 @@ impl From<StorageFormatScanReport> for StorageCompatibilityReport {
 /// Pure report builder shared by the command and its tests: `scan_profiles`
```
daopunk: This is no longer a pure report builder — it may `rename` profile directories. Please rewrite the docstring to say that.

The `compatibility_report_*` / `get_storage_compatibility_command_*` tests lock this file’s `ENV_TEST_MUTEX` but do not `EnvGuard::unset("PACTO_TEST_SANDBOX_ROOT")`. `test_sandbox.rs` and `sandbox_handle.rs` serialize the same process-global env with different mutexes; that flake is worse now that this function is destructive. Unset the env in every non-doctor `compatibility_report` test.

src-tauri/src/storage_format.rs
```diff
@@ -233,6 +242,215 @@ pub(crate) fn scan_profiles(app_data_dir: &Path) -> StorageFormatScanReport {
     }
 }
 
+// -- storage doctor / quarantine (U15) -----------------------------------
```
daopunk: Repo rule: no spec-checklist breadcrumbs in `src/`. Drop `(U15)` here and on the test-module comment below; the plan unit id belongs in `docs/plans/…` only.

daopunk (commented):
## Review

Approve with comments. The doctor matches U15: same Unrecognized/Divergent verdicts, sandbox-root-only moves, symlink re-check before rename, timestamp+sequence destinations, no new Tauri command. Production `get_storage_compatibility` cannot aim this at the real OS account (`<root>/data` is never `…/io.pacto`). CI is green.

Do not merge until the real-OS recovery runbook is restored in `OPERATOR_UPDATES.md` — the block screen still hides which npub is at fault, and quarantine does not run outside a sandbox. Other notes are defense-in-depth / hygiene.

Inline comments cover the four items.

src-tauri/src/storage_format.rs
logicminds: > Repo rule: no spec-checklist breadcrumbs in `src/`. Drop `(U15)` here and on the test-module comment below; the plan unit id belongs in `docs/plans/…` only.

Addressed in f65c973 — dropped `(U15)` from both section-divider comments.

src-tauri/src/storage_format.rs
logicminds: > Refuse the whole scan unless `app_data_dir` itself canonicalizes under the sandbox root (ideally `<root>/data`) before classifying anything. Add a test: env set, scanned directory *outside* that root, nothing moved.

Addressed in f65c973 — `quarantine_stale_profiles` now canonicalizes and checks `app_data_dir` against the sandbox root up front and returns before touching `read_dir` if it doesn't match, instead of relying on N per-profile `revalidate_before_move` refusals. Added `quarantine_refuses_a_scan_directory_outside_the_sandbox_root` covering exactly that case.

The separate `cfg!(debug_assertions) && sandbox_root().is_some()` restructuring you flagged as belt-and-suspenders was explicitly marked "not required for correctness" and is superseded by the already-landed widened identifier gate, so left as-is.

docs/build/OPERATOR_UPDATES.md
logicminds: > Please restore the runbook under a heading for the real OS data directory (keep the new sandbox auto-recovery as its own subsection). Also: the gate is the env var, not git branch `main`.

Addressed in f65c973 — restored the manual sqlite3 runbook under its own "Manual recovery for the real OS data directory" heading, and reworded the automatic-recovery section to state the actual gate (`PACTO_TEST_SANDBOX_ROOT` or a non-`io.pacto` running identifier) instead of framing it as "on `main`".

PR #264, opened 2026-08-13
fix: refuse a second launch against one sandbox root

logicminds:
## Summary

Two follow-ups from the PR #261 review, both consequences of skipping the Tauri single-instance plugin for dev sandboxes. That skip is what finally let two sandboxes run side by side - but it also removed the guard that incidentally prevented two launches against the *same* root.

- **Before:** a second launch against one sandbox root silently overwrote the first's handle while both processes ran against the same SQLite and MLS store. Concurrent sandboxes also restored to identical stacked window geometry, landing on top of each other.
- **After:** the second launch refuses and names the holder; a root whose holder is gone is reclaimed rather than wedged; each sandbox restores its own geometry.

## Changes

- **Per-root lockfile** in `test_sandbox`, making the invariant structural instead of caller discipline. `strictPort` only ever protected the `make dev` paths.
- Staleness reasoning is ported from the port-claim system already proven in `scripts/dev-ports.mjs` - `O_EXCL` create, pid liveness where `EPERM` counts as alive, and a grace window - rather than inventing a second, differently-behaved notion of "stale".
- **Window-state filename** keyed off the sandbox root, so concurrent sandboxes no longer share one geometry file.
- Both are debug-only by construction, mirroring `multi_instance_allowed`'s existing posture; a release build has no sandbox concept to honor.
- `docs/build/DEV_SANDBOX.md` previously documented the same-root gap as a known limitation; that text now describes the guard.

## Test plan

- `cargo test --lib test_sandbox` - **24 passed** (8 pre-existing + 16 new).
- New coverage: refusal naming a live holder; reclaim of a dead holder (spawns and reaps a real child process, backdating the record past the grace window); clean release-then-relaunch with no leaked lockfile; **two different roots still both launch**, so this does not regress what #261 fixed; a plain dev run with no sandbox root is unaffected; debug-only posture asserted the same way `multi_instance_allowed`'s own test does; window-state filenames diverge across roots and never equal the plugin default when a root is set.
- `cargo test --lib` - 663 passed, 0 failed. Clippy at baseline (one new warning was introduced and fixed in review).

---

### DIFF
--- a/src-tauri/src/test_sandbox.rs
+++ b/src-tauri/src/test_sandbox.rs
@@ -1,3 +1,4 @@
+use serde::{Deserialize, Serialize};
 use std::path::{Component, Path, PathBuf};
 use tauri::{AppHandle, Manager, Runtime};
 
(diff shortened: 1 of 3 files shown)


src-tauri/src/test_sandbox.rs
```diff
@@ -156,6 +157,211 @@ pub fn enforce_dev_world_root() -> Result<(), String> {
...
+    std::io::Error::last_os_error().raw_os_error() == Some(libc::EPERM)
+}
+
+#[cfg(not(unix))]
+fn pid_is_alive(pid: u32) -> bool {
+    // No `kill(pid, 0)`-equivalent probe without an extra dependency on this
+    // platform. Treat a recorded holder as alive unless it names this very
+    // process, so an unsupported platform fails safe (refuses) instead of
+    // guessing at a stranger process's liveness.
+    pid == std::process::id()
```
logicminds: The comment says fail closed: treat a recorded holder as **alive unless it names this process**. The comparison does the opposite.

On Windows a second debug launch against the same root sees the first instance's pid, `pid == std::process::id()` is false, `pid_is_alive` is false, and after `LOCK_STALE_GRACE_MS` (1s) `lock_is_stale` is true. The second process unlinks `sandbox.lock` and proceeds. Both then mutate the same SQLite/MLS store — the corruption this PR exists to prevent.

`scripts/dev-ports.mjs` `isPidAlive` has no such inversion: `process.kill(pid, 0)` / EPERM means alive on every platform Node supports.

```rust
pid != std::process::id()
```

That refuses a foreign holder forever on platforms without a probe (manual `sandbox.lock` delete remains the recovery, as DEV_SANDBOX.md already documents). The existing live-holder test uses `std::process::id()`, so it stays green even with the inverted comparison; add a planted foreign pid past the grace window (or a unix `sleep` child) so this cannot regress.

src-tauri/src/test_sandbox.rs
```diff
@@ -399,4 +605,148 @@ mod tests {
...
+    }
+
+    #[test]
+    fn sandbox_launch_lock_refuses_when_holder_is_live_and_names_it() {
+        let _lock = env_lock();
+        let dir = temp_test_dir("launch-lock-live-holder");
+        let _root = EnvGuard::set(SANDBOX_ROOT_VAR, dir.to_str().unwrap());
+
+        let canonical = dir.canonicalize().unwrap();
+        let holder_pid = std::process::id();
```
logicminds, line 657: This is the only live-holder coverage, and it records **this test process's pid**. On `cfg(not(unix))` that is the unique pid `pid_is_alive` still treats as alive, so the test cannot catch the `==` / `!=` inversion.

A unix child (`sleep` without waiting) as the recorded holder, plus a `cfg(not(unix))` case that plants a non-self pid older than `LOCK_STALE_GRACE_MS` and expects refuse, would lock the fail-closed contract in.

logicminds (commented):
## Review summary

The unix lock path does what the PR promises: `create_new` for exclusive claim, `kill(pid, 0)` with EPERM counted as alive, named refusal, Drop-on-exit so a clean relaunch does not wait out the grace window, and a sandbox-keyed window-state filename so concurrent roots do not stack.

One logic bug on the non-unix fallback should be fixed before merge (inline below). I applied it locally as `fix(review): fail closed on non-unix sandbox lock liveness` (`20dfb9a`) with a live-foreign-pid test; it is not pushed.

Otherwise this looks ready: debug-only like `multi_instance_allowed`, no-op without a sandbox root, different roots still launch side by side.

**Suggested change:** invert `pid_is_alive` on `cfg(not(unix))` from `pid == std::process::id()` to `pid != std::process::id()`, matching the comment. A foreign pid must be treated as live so a second Windows debug launch cannot reclaim a live first instance after the 1s grace window and share its SQLite/MLS store.

src-tauri/src/test_sandbox.rs
logicminds: > The comparison does the opposite.

Addressed: inverted the non-unix `pid_is_alive` check to `pid != std::process::id()`, so a foreign holder is treated as live and refused instead of reclaimed after the 1s grace window. Manual `sandbox.lock` delete remains the recovery path on platforms without a probe.

PR #267, opened 2026-08-14
feat: offer a refusable join for groups that added you without an invite dm

logicminds:
> **Stacked on #268**, which repairs the unparseable `ci.yaml` that had been failing every run at startup. Without it this PR gets no CI at all. GitHub retargets this to `main` automatically once #268 merges.

## Summary

A group creator can add you over MLS alone — KeyPackage to Welcome, with no `squad_invite` DM. The engine accepted that welcome and `list_pending_mls_welcomes` reported it, but squads are materialized from a DM invite payload, so nothing ever offered the join and the group stayed invisible.

It was worse than invisible. `list_pending_mls_welcomes` already resolves the inviter's display name, fires an **OS notification** and writes a **durable Catch up row**. So the app announced an invite and then rendered a row that could not be clicked — permanently, across restarts.

**Before:** a bot or non-Pacto client adds you to a group; you get a desktop notification, a Catch up row with a disabled control beside a bare ellipsis, and no way to join, refuse, or dismiss it. Ever.

**After:** the invitation appears under **DMs → Requests** with the group's real name, description, image, inviter and member count. **Join** materializes the squad through the same path a DM invite uses. **Ignore** hides it locally and clears the Catch up row.

## Changes

- **`src/lib/invites/pending-welcomes.ts`** — pure `offeredWelcomes()` filter. Drops groups already joined, refused, blocked, in-flight, or awaiting a consent-first admission, with one trim/lowercase normalization path for every id filter.
- **`pending-welcomes-store.ts`** — the derived list plus `declineWelcomeForGroup`.
- **`PendingWelcomeRequests.svelte`** — new, runes mode. Cards above the Requests list. The legacy `MessengerNavbar` shell stays legacy and hosts it as a child; it also suppresses its "No requests" empty state when cards are present.
- `declinedWelcomeGroupIds` is npub-scoped and cleared with account state. `finalizeSquadAfterAnnouncementsWelcome` now takes `messageId: string | null`, since no DM backs a bare welcome.
- Declining a `squad_invite` DM records its group id too, so both surfaces agree on one refusal.
- Catch up's welcome rows are repaired with the same reliable source: the real group name, opening DMs → Requests.
- Removed a stale comment referring to an "unattributed Add to squad modal". Traced through history — that modal never existed.

### Two decisions worth a reviewer's attention

**Attribution is deliberately not attempted.** `findWelcomeInviteSource` scans the DM cache, and `init_finished` loads only *one message per chat*, so it reports "no invite DM" for plenty of genuinely invited groups — that is the actual cause of the broken Catch up row. DM content is encrypted at rest, so the backend cannot cheaply answer either. Rather than guess, **every** unresolved pending welcome is offered. That is safe because the squad flow is consent-first: the admitter adds you only after you accept, which is what `pendingSquadAdmissions` records, so a pending welcome normally means its DM invite is already spoken for. An invite-first sender is precisely this surface's audience.

**Refusal is local only, and the UI says so.** MLS has no decline primitive — only `accept_mls_welcome` exists. Ignoring hides the card; the welcome stays pending in the engine and the inviter is told nothing. The card states this rather than implying otherwise. The trust posture is unchanged in the direction that matters: anyone who can resolve your KeyPackage can put a *prompt* in your Requests tab, which is the same property inbound DMs already have, and it inherits the existing `blockedDmNpubs` blocklist.

## Test plan

- `pnpm check` — 0 errors. `pnpm lint` — 0 errors (1 pre-existing warning in an unrelated dialog harness). `pnpm test` — 1976/1976.
- 28 new unit tests: `pending-welcomes.test.ts` (21) and `resolve-welcome-entry.test.ts` (7). The case/whitespace-insensitivity tests were falsified on purpose — breaking the normalization turns exactly those red.

## Evidence

Driven through a real local stack: `make dev-sandbox` pointed only at `wss://localhost:7001`, with bot `bosun` creating MLS groups that invite the sandbox identity and send **no DM**.

| Checked | Observed |
|---|---|
| Card content | "Bare Welcome Probe", inviter npub, `2 members` — real metadata, correct plural |
| Join | squad materialized with default channels, navigated to Squads, ready toast |
| Catch up row after Join | resolved backend-side by `do_accept_mls_welcome`; only an unrelated `action_prompt` remained |
| Ignore | welcome **still pending in the engine**, group **not** joined, Catch up row cleared |
| Already-joined group | filtered out of the list live |
| After restart | refused welcome still pending in the engine but **not re-offered**, proving the npub-scoped load path |

## Known limitation

A bare welcome for a channel-in-squad group is indistinguishable from a squad welcome at the protocol level, so accepting one would materialize a squad named after the channel group. It requires a direct add with no DM, and is not corrupting. Recorded on `pacto-app-384.70` rather than papered over.

---


src/lib/invites/pending-welcomes.ts
```diff
@@ -0,0 +1,124 @@
...
+  }
+  return offered;
+}
+
+/**
+ * Join the group behind a pending welcome: accept it in the engine, then run the
+ * same squad materialization the DM invite path runs. No DM message backs this,
+ * so no invite-decision id is recorded.
+ */
+export async function acceptOfferedWelcome(welcome: OfferedWelcome): Promise<void> {
```
logicminds, line 160: **P1 #1 (reliability, adversarial -- confidence 100)** -- No partial-failure recovery: engine-accept-succeeds-but-finalize-fails re-strands the exact bug this PR fixes

MLS has no decline primitive -- once `acceptMlsWelcome` succeeds the welcome is consumed at the engine level (confirmed via `do_accept_mls_welcome` in `src-tauri/src/lib.rs`). If the following `finalizeSquadAfterAnnouncementsWelcome(...)` call throws (it already has a rollback path for `persistSquad` failure, shared with the DM-invite caller), the group is accepted by the engine but never materialized locally -- reproducing the exact "permanently invisible group" bug this PR sets out to fix. The card's own comment two lines up in `PendingWelcomeRequests.svelte` ("the welcome is still pending in the engine, so a retry is legitimate") is then false: retry re-calls `acceptMlsWelcome` on a welcome the engine no longer reports as pending and fails identically.

**Suggested fix:** split the two awaits so callers can distinguish "engine accept failed, safe to retry" from "engine accepted, only local materialization failed." On the latter, persist a durable "accepted-but-not-materialized" record (mirroring `pendingSquadAdmissions`/`tryCompletePendingSquadAdmission`) keyed by `groupId`, surfaced independently of `pendingMlsWelcomes` membership, and route retry to re-run only `finalizeSquadAfterAnnouncementsWelcome` -- never re-call `acceptMlsWelcome`.

logicminds, line 160: **P1 #2 (testing -- confidence 100)** -- `acceptOfferedWelcome` (the sole `messageId: null` caller) has zero test coverage

`acceptOfferedWelcome` is the only exercise of the new `messageId: string | null` contract on `finalizeSquadAfterAnnouncementsWelcome`, and it is never imported or called by any test -- `pending-welcomes.test.ts` only covers the pure `offeredWelcomes()` filter.

**Suggested fix:** add tests mocking `acceptMlsWelcome`/squad persistence (mirroring `accept-invite.test.ts`) that call `acceptOfferedWelcome` and assert: the squad is added, `acceptedSquadInviteIds` is unchanged (no `null`/`"null"` entry), and a mocked `acceptMlsWelcome` rejection propagates. Also add coverage for the partial-failure path from #1 once it's fixed.

src/lib/invites/pending-welcomes.ts
```diff
@@ -0,0 +1,124 @@
...
+  joiningGroupIds,
+}: OfferedWelcomeInputs): OfferedWelcome[] {
+  // One normalization path for every id filter: group ids reach us from MLS
+  // metadata, squad rows and localStorage, and their case does not always match.
+  const resolved = new Set(
+    [
+      ...squadIds,
+      ...declinedGroupIds,
+      ...pendingAdmissionGroupIds,
+      ...joiningGroupIds,
```
logicminds: **P1 #3 (correctness, julik-frontend-races -- confidence 100)** -- "Joining" filter unmounts the card, making the disabled 'Joining...' state unreachable

`offeredWelcomes()` folds `joiningGroupIds` into the same `resolved` exclusion `Set` as `squadIds`/`declinedGroupIds`/`pendingAdmissionGroupIds`. `PendingWelcomeRequests.svelte` writes `joiningWelcomeGroupIds` synchronously before `await acceptOfferedWelcome(welcome)`, so the card's own group id is immediately filtered out of `$offeredWelcomeList` and its `{#each}` row unmounts -- before the `{@const joining}`-driven disabled `Joining...` button state it computes can ever render. Double-submit protection is already provided by the `$joiningWelcomeGroupIds.includes(...)` guard and `disabled={joining}`; this list-level exclusion is redundant for that purpose and defeats the intended UX (the component's own comment: "leaving the card in place is the useful failure").

**Suggested fix:** drop the `...joiningGroupIds` spread from the `resolved` set here (and remove `joiningGroupIds` from `OfferedWelcomeInputs` / the derived store's dependency wiring in `pending-welcomes-store.ts`). Update the "filters out a welcome whose group has an accept already in flight" test accordingly and add a component-level assertion that Join shows the disabled `Joining...` label mid-accept.

src/components/dm/PendingWelcomeRequests.svelte
```diff
@@ -0,0 +1,235 @@
...
+      failedGroupId = welcome.groupId;
+    } finally {
+      joiningWelcomeGroupIds.update((ids) => ids.filter((id) => id !== welcome.groupId));
+    }
+  }
+</script>
+
+{#if $offeredWelcomeList.length > 0}
+  <ul class="welcome-list" role="list" aria-label={$t('messaging.pendingWelcome.sectionLabel')}>
+    {#each $offeredWelcomeList as welcome (welcome.groupId)}
```
logicminds, line 47: **P1 #4 (adversarial -- confidence 75)** -- Bare-welcome spam has no cap -- unbounded cards can bury real DM requests

`offeredWelcomeList` has no cap/limit and this renders one `<li>` per entry with no slice/virtualization, unconditionally mounted above the real Requests list. Anyone who can resolve the victim's MLS KeyPackage can push unlimited bare welcomes with distinct group ids; only the pre-existing per-npub `blockedDmNpubs` list mitigates this.

**Suggested fix:** cap `offeredWelcomeList` (e.g. slice to most-recent N, with pagination) or paginate the render here; consider collapsing multiple distinct group ids from the same welcomer npub into one entry.

src/components/dm/PendingWelcomeRequests.svelte
```diff
@@ -0,0 +1,235 @@
...
+      {@const joining = $joiningWelcomeGroupIds.includes(welcome.groupId)}
+      <li class="welcome-card">
+        <div class="welcome-icon">
+          {#if welcome.imageUrl}
+            <img src={welcome.imageUrl} alt="" class="welcome-icon-img" />
```
logicminds: **P1 #5 (security -- confidence 75)** -- Unvalidated `group_image_url` auto-loads as `` from an unvetted sender (tracking-pixel / IP exposure)

`welcome.imageUrl` comes straight from `welcome.group_image_url?.trim()` with no scheme/host validation and loads automatically for every unresolved welcome, including from senders the user has never interacted with -- a classic tracking-pixel / IP-deanonymization vector, notable given Pacto's privacy/censorship-resistance positioning. This is a new instance of a pre-existing pattern (`getProfileAvatarSrc` has the same gap for DM avatars), but this is the first surface that auto-shows metadata from a genuinely unvetted party.

**Suggested fix:** add `referrerpolicy="no-referrer"` (and `loading="lazy"`) here as a partial mitigation. Full fix is systemic (proxy remote image fetches through a trusted backend endpoint, or gate behind explicit opt-in for unvetted senders) and likely out of this PR's scope alone -- tracked as a residual risk.

src/lib/invites/pending-welcomes-store.ts
```diff
@@ -0,0 +1,61 @@
...
+import { resolveOneCatchUpEntry } from '../../stores/catch-up';
+import { sameMlsGroupId } from './accept-invite';
+import {
+  offeredWelcomes,
+  recordDeclinedWelcomeGroupId,
+  type OfferedWelcome,
+} from './pending-welcomes';
+
+/** Group ids with an accept in flight, so a card cannot be double-submitted. */
+export const joiningWelcomeGroupIds = writable<string[]>([]);
```
logicminds, line 27: **P1 #6 (adversarial -- confidence 75)** -- `joiningWelcomeGroupIds` is not reset by `clearAccountState` -- an in-flight join can leak into the next account

Unlike its sibling `declinedWelcomeGroupIds`, this new store is never referenced from `clear-account-state.ts`. `finalizeSquadAfterAnnouncementsWelcome`'s `squads.update`/`persistSquad` calls have no active-npub check, so an in-flight `acceptOfferedWelcome` promise started before a logout/account-switch can still mutate the live `squads` store and persist to whatever backend session is active when it resolves.

**Suggested fix:** add `joiningWelcomeGroupIds.set([])` to `clearAccountState`, and capture/re-check the active npub before the store mutations in `acceptOfferedWelcome`/`finalizeSquadAfterAnnouncementsWelcome`, aborting if it changed mid-flight.

src/components/dm/PendingWelcomeRequests.svelte
```diff
@@ -0,0 +1,235 @@
...
+            <p class="welcome-error" aria-live="polite">
+              {$t('messaging.pendingWelcome.joinFailed')}
+            </p>
+          {/if}
+          <div class="welcome-actions">
+            <button
+              type="button"
+              class="welcome-btn welcome-btn-accept"
+              disabled={joining}
+              onclick={() => accept(welcome)}
```
logicminds: **P2 #7 (project-standards -- confidence 100)** -- `onclick={() => accept(welcome)}` omits the `void` prefix `AGENTS.md`/`CLAUDE.md` requires for fire-and-forget async

The repo's async convention ("Fire-and-forget async work is prefixed with `void`") is already followed elsewhere -- `MlsMembersPanel.svelte:114` uses `onclick={() => void confirmRestore()}`.

**Suggested fix:** `onclick={() => void accept(welcome)}`.

logicminds (commented):
## Code Review (ce-code-review)

Automated multi-persona review: correctness, project-standards, testing, maintainability, security, reliability, julik-frontend-races, adversarial (in-process fallback -- no cross-model peer CLI available on this host). All 10 findings below were independently re-validated against the live workspace (including the Rust `do_accept_mls_welcome` path for #1) -- 10/10 confirmed, 0 dropped.

**Verdict: Ready with fixes.** No P0s. Two P1s directly undercut this PR's own stated purpose and share one fix location (`pending-welcomes.ts:67-108`):
- **#1** -- if `finalizeSquadAfterAnnouncementsWelcome` throws after `acceptMlsWelcome` already succeeded, the group is stuck invisible again (the MLS engine has no re-offer primitive) -- reproducing the exact bug this PR fixes. The card's "retry is legitimate" comment is then false.
- **#3** -- `offeredWelcomes()` folds `joiningGroupIds` into its exclusion set, so the card unmounts the instant `accept()` starts, making the disabled "Joining..." state dead code.

#4 (unbounded bare-welcome spam), #5 (unvalidated remote image URL -- tracking-pixel exposure from an unvetted sender), and #6 (`joiningWelcomeGroupIds` not cleared on logout) are real hardening gaps worth closing before wide rollout. #2/#7/#8/#9/#10 are smaller (test coverage, a `void`-prefix convention nit, a swallowed-rejection bug, and a DM-supplied-id trust gap).

**Fix order:** #3 -> #1 -> #2 (tests for both) -> #10 -> #9 -> #8 (decline-path group) -> #6, #5, #4 (hardening) -> #7 (nit).

Full report with reasoning, evidence, and residual risks/testing gaps: see inline comments below.

src/lib/invites/pending-welcomes-store.ts
```diff
@@ -0,0 +1,61 @@
...
+);
+
+/**
+ * Refuse a pending welcome and clear the Catch up row that announced it.
+ *
+ * Accepting needs no equivalent: `do_accept_mls_welcome` resolves the entry
+ * backend-side. Refusing leaves the welcome pending in the engine, so without
+ * this the row would survive and point at a Requests tab no longer offering it.
+ */
+export function declineWelcomeForGroup(groupId: string): void {
```
logicminds, line 60: **P2 #8 (testing -- confidence 100)** -- `declineWelcomeForGroup` has zero test coverage (both branches unverified)

No test file exists for `pending-welcomes-store.ts`; neither the "no matching pending welcome" early return nor the swallowed-rejection branch below (#9) is exercised.

**Suggested fix:** add `pending-welcomes-store.test.ts` covering: records the decline even with no matching `pendingMlsWelcomes` entry (and skips `resolveOneCatchUpEntry` then); calls `resolveOneCatchUpEntry` with `wrapper_event_id` when a match exists (including case/whitespace-differing group ids); does not throw when `resolveOneCatchUpEntry` rejects.

src/lib/invites/pending-welcomes-store.ts
```diff
@@ -0,0 +1,61 @@
...
+  recordDeclinedWelcomeGroupId(groupId);
+  const pending = get(pendingMlsWelcomes).find((w) => sameMlsGroupId(w.nostr_group_id, groupId));
+  if (!pending) return;
+  void resolveOneCatchUpEntry(pending.wrapper_event_id).catch(() => {
```
logicminds: **P2 #9 (reliability -- confidence 75)** -- Swallowed `resolveOneCatchUpEntry` rejection leaves the Catch up row stuck -- the "next refresh clears it" comment is factually wrong

`resolveOneCatchUpEntry` (in `stores/catch-up.ts`) awaits the backend call before touching local state, and `hydrateCatchUp` does no client-side reconciliation against `declinedWelcomeGroupIds` -- so a failed resolve here leaves a dead Catch up row that a "next refresh" will not clear, and by then `offeredWelcomeList` has already filtered the group out via `declinedGroupIds`, so clicking the stuck row opens an empty Requests tab.

**Suggested fix:** retry `resolveOneCatchUpEntry` (e.g. re-attempt on next Catch up hydrate for failed `wrapper_event_id`s) or at minimum log via `dmError` instead of an empty catch, and fix the comment.

src/routes/+page.svelte
```diff
@@ -1019,6 +1021,10 @@ import MyDashboard from '../components/parent/MyDashboard.svelte';
                 onAcceptChannelInSquad={acceptChannelInSquadInvite}
                 onDeclineSquad={(msg: DmMessage) => {
                   declinedSquadInviteIds.update((ids: string[]) => (ids.includes(msg.id) ? ids : [...ids, msg.id]));
+                  // Also refuse the matching MLS welcome, so a pending-welcome
+                  // card cannot re-offer a squad the user just turned down.
+                  const declined = parseSquadInviteMessage(msg.content ?? '');
```
logicminds, line 1026: **P2 #10 (adversarial -- confidence 75)** -- DM-invite decline records an unverified, DM-self-reported `groupId` -- can gate a different real bare-welcome card

`parseSquadInviteMessage(msg.content)` parses an untrusted DM payload's self-reported `groupId` and calls `declineWelcomeForGroup(declined.groupId)` unconditionally, with no check that the id matches a real `$pendingMlsWelcomes` entry. A DM-declared `groupId` with no cryptographic binding to any real MLS group can be recorded as declined while a same-named but differently-keyed real bare-welcome card remains fully live.

**Suggested fix:** only record the decline when `declined.groupId` matches an entry in `$pendingMlsWelcomes` -- otherwise the DM's self-reported id is unverifiable and should not silently gate a different, real card.

src/lib/invites/pending-welcomes.ts
logicminds: > persist a durable "accepted-but-not-materialized" record ... and route retry to re-run only `finalizeSquadAfterAnnouncementsWelcome`

Addressed in 6897e26 — after `acceptMlsWelcome` succeeds we persist a npub-scoped `pendingWelcomeFinalizations` row (same pattern as `pendingSquadAdmissions`) before finalize. Retry and login hydrate re-run only finalize, never `acceptMlsWelcome`.

logicminds: > `acceptOfferedWelcome` ... is never imported or called by any test

Addressed in 6897e26 — `pending-welcomes.test.ts` now covers the happy path (`finalize` called with `null`), accept rejection, persist failure plus retry skipping `acceptMlsWelcome`, and account-switch abort.

src/lib/invites/pending-welcomes.ts
logicminds: > drop the `...joiningGroupIds` spread from the `resolved` set

Addressed in 6897e26 — `joiningGroupIds` is gone from the filter and the derived store, so the card stays mounted and Joining… can render. Covered by `PendingWelcomeRequests.test.ts`.

src/components/dm/PendingWelcomeRequests.svelte
logicminds: > cap `offeredWelcomeList` (e.g. slice to most-recent N, with pagination) or paginate the render here; consider collapsing multiple distinct group ids from the same welcomer npub into one entry.

Good catch on the spam surface. This is a product call more than a one-line fix — a hard cap hides later real welcomes, pagination needs UI, and collapsing by welcomer changes the card model. Going to think through cap vs paginate vs collapse-by-npub vs treating KeyPackage as the trust boundary before picking one.

src/components/dm/PendingWelcomeRequests.svelte
logicminds: > add `referrerpolicy="no-referrer"` (and `loading="lazy"`) here as a partial mitigation

Addressed in 6897e26 — the welcome icon now sets `referrerpolicy="no-referrer"`, `loading="lazy"`, and `decoding="async"`. Full proxy/opt-in is still out of scope for this PR.

src/lib/invites/pending-welcomes-store.ts
logicminds: > add `joiningWelcomeGroupIds.set([])` to `clearAccountState`, and capture/re-check the active npub

Addressed in 6897e26 — `clearAccountState` now resets `joiningWelcomeGroupIds` and the new finalization store. `acceptOfferedWelcome` captures `currentNpubForPersistence` and aborts live store mutations if it changed; I did not widen `finalizeSquadAfterAnnouncementsWelcome` itself (shared DM-invite path this PR did not introduce).

src/components/dm/PendingWelcomeRequests.svelte
logicminds: > `onclick={() => accept(welcome)}` omits the `void` prefix

Addressed in 6897e26 — `onclick={() => void accept(welcome)}`.

src/lib/invites/pending-welcomes-store.ts
logicminds: > add `pending-welcomes-store.test.ts` covering: records the decline even with no matching `pendingMlsWelcomes` entry ... does not throw when `resolveOneCatchUpEntry` rejects.

Addressed in 6897e26 — new `pending-welcomes-store.test.ts` covers the no-match skip, `wrapper_event_id` resolve including case/whitespace, swallowed rejection, and clearing a stuck finalization row.

src/lib/invites/pending-welcomes-store.ts
logicminds: > retry `resolveOneCatchUpEntry` ... or at minimum log via `dmError` instead of an empty catch, and fix the comment.

Addressed in 6897e26 — one retry, then `dmError`. The comment now states that `hydrateCatchUp` does not reconcile against `declinedWelcomeGroupIds`. Durable retry-on-hydrate is left for a follow-up.

src/routes/+page.svelte
logicminds: > only record the decline when `declined.groupId` matches an entry in `$pendingMlsWelcomes`

Not addressing: unmatched DM `groupId`s cannot filter a different welcome — `offeredWelcomes` matches on group id, not name. Recording the decline even when the welcome has not arrived yet is the point of that call site, so a later bare-welcome card cannot re-offer a squad the user just turned down. Gating the record on a currently-pending welcome would reopen that race.

src/components/dm/PendingWelcomeRequests.svelte
logicminds: > cap `offeredWelcomeList` (e.g. slice to most-recent N, with pagination)

Addressed in 42ccb56 — pending-welcome cards are capped at the 20 newest. Engine-accepted recovery rows stay visible so a persistSquad retry cannot be hidden by the cap.

PR #105, opened 2026-07-18
feat(nostr): relay connection diagnostics, pre-add probe, and TLS certificate inspection

logicminds:
## Summary

This PR makes Nostr relay connection problems discoverable and actionable from the settings UI.

Before, a failing relay only showed a generic status label. Users had to save a custom relay before discovering it was unreachable, slow, expired, or not a relay, and TLS certificate details were completely hidden.

After, users can probe a relay before adding it, see a classified failure reason next to disconnected relays, expand any relay to inspect recent logs, byte/event metrics, and TLS certificate metadata, and get warned about certificates that are expired or expiring soon.

## Changes

- Backend (`src-tauri/src/lib.rs`):
  - Store per-relay failure reasons, redacting URL-like tokens so secrets stay out of the UI.
  - Classify transport errors into stable categories: DNS lookup failed, connection refused, timeout, network unreachable, TLS handshake/expired, protocol error, etc.
  - Add `probe_relay` for a throwaway pre-add health check and `get_relay_certificate` to read TLS certificate metadata (subject, issuer, SANs, fingerprint, key algorithm, validity window).
  - Extend `get_relays` to return `failure_reason` for non-connected relays.
- Frontend (`src/components/settings/NostrSettingsSection.svelte`, `src/lib/api/relays.ts`):
  - Add a pre-add probe prompt with status, RTT, and actionable messages.
  - Add a per-relay expandable diagnostics panel with logs, metrics, and certificate summary.
  - Add typed DTOs and command wrappers for the new backend commands.

## Test plan

- `cd src-tauri && cargo test relay` — 20 tests passed.
- `pnpm test` — 151 test files / 1,275 tests passed.
- `pnpm check` — 0 errors (pre-existing warnings only).

## Post-Deploy Monitoring & Validation

- In Settings → Nostr Relays, verify that known-good relays show "Healthy" when probed and that unreachable relays surface classified reasons instead of just "Disconnected".
- Verify TLS certificate inspection populates for `wss://` relays and flags expired or near-expiry certificates.
- Confirm failure reasons and probe messages do not contain secrets, auth tokens, or query parameters.
- Watch Sentry / logs for increases in `probe_relay` or `get_relay_certificate` errors.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

PR #270, opened 2026-08-15
feat: relay connection diagnostics, pre-add probe, and TLS certificate inspection

logicminds:
## Summary

A disconnected relay used to look identical whether the cause was DNS, a bad certificate, or a wrong port, and there was no way to check a candidate relay before saving it. This adds three diagnostics to the existing Settings > Nostr relay list and detail panel:

- **Classified failure reason** — a relay that failed to connect shows *why* (DNS lookup failed, connection refused, timeout, TLS failure, etc.) beside its status. Never shown for a connected relay; clears on reconnect, removal, disable, and logout.
- **Pre-add probe** — test a candidate relay URL before saving it. Resolves, connects, and runs one read-only query without ever joining the live relay pool or writing anything to the candidate relay. The operator can still add a relay that fails the probe.
- **TLS certificate inspection** — expanding a `wss://` relay's detail panel shows its certificate (subject, issuer, validity window, SAN, SHA-256 fingerprint, key algorithm) with a computed expiry verdict, and a persistent disclosure that trust was never validated — the panel can never be read as a trust verdict.

Supersedes draft PR #105: this rebuilds the same surface against the current detail panel (its frontend predates it and isn't portable), replaces message-text error classification with a typed downcast, replaces hardcoded English DTOs with machine codes the frontend labels, and replaces a standard trust-checking `ClientConfig` (which aborts on exactly the expired certificate this feature exists to warn about) with a capture-only verifier structurally confined to its own module.

Executes `docs/plans/2026-08-14-001-feat-relay-diagnostics-plan.md` (beads epic `pacto-app-4c0`, 8 implementation units).

## Screenshots

Classified failure reason next to a failed relay, and a healthy relay for contrast:

Classified failure reason on the relay list

Pre-add probe result before saving a candidate relay:

Pre-add probe result

TLS certificate panel with the trust-not-evaluated disclosure, against a real relay's live certificate:

TLS certificate panel

## Backend

- `classify_relay_error`: typed downcast through `TransportError::Backend` → `async_wsocket::Error` → `tungstenite::Error`, mapping to a closed set of stable codes. No message-text parsing.
- `redact_urls_in_text` extended to `ws://`/`wss://`: drops the entire query string for relay schemes rather than filtering by the existing 9-name allowlist built for RPC URLs.
- `probe_relay`: throwaway `RelayPool`, single 10s deadline covering DNS + connect + one read-only query, torn down on every exit path. Never touches the app's live client.
- `relay_cert.rs`: DER parsing, a clock-injectable expiry verdict (single 30-day-threshold authority), and an isolated TLS handshake behind a capture-only `ServerCertVerifier`. The verifier and its `ClientConfig` are private to this module — containment is structural, enforced by a source-scan test that fails if a second construction site appears anywhere in `src-tauri/src`.
- Login-generation guard so a stale monitor loop (from a previous account, in the same process) can't attribute diagnostics to the current one; logout clears failure reasons, cached certificates, and the pre-existing relay logs/metrics statics that had no clear site before.
- No SQLite migration. No new crypto-provider feature — `tokio-rustls`/`aws-lc-rs` already resolved via the existing `rustls` dependency.

## Frontend

- `src/lib/api/relays.ts`: typed DTOs, `probeRelay`/`getRelayCertificate` wrappers, and label mappers with a fallback for unrecognized codes.
- `NostrSettingsSection.svelte`: failure reason on the relay row, a probe control on the add-relay form (disabled while in flight or the URL is invalid, result announced via `aria-live`), and a certificate block gated to `wss://` relays with its own expiry-state color palette (never reusing the muted "manually disabled" status color for an expired certificate).
- `en`/`es` i18n catalogs at exact key parity.

## Found and fixed along the way

Driving the new detail panel live against a real local relay surfaced a pre-existing crash: the relay activity log's `{#each}` block keyed on `timestamp + message`, which collides when the reconnect loop logs the same status-change message twice within the same second (now more likely, since U2 forces a disconnect before more reconnect attempts than before). Fixed to key on index.

## Verification

- `cd src-tauri && cargo test --lib` — 734 passed (run 4× for the parallelism-sensitive diagnostics-clearing tests)
- `pnpm test` — 2001 passed
- `pnpm check`, `pnpm lint` — clean
- `pnpm check:tauri-commands` — no new orphaned commands; `probe_relay` and `get_relay_certificate` both have real callers
- `cargo tree -e features -i rustls` — no new crypto-provider feature (confirmed via `Cargo.lock` diff: only `rcgen`/`x509-parser` are new packages)
- `local-relay-tls`, the `tokio-tungstenite` block, and `rustls = "0.23"` are byte-for-byte unchanged
- Driven live via the Tauri MCP bridge against a real local relay (`wss://localhost:7001`, real mkcert certificate): classified failure reasons rendering from the actual reconnect loop, the full probe lifecycle (disabled → in-flight → `dns_failed`/`connection_refused`/`not_a_relay`, and adding a relay after a failed probe), the certificate panel with a real handshake and a `Valid` verdict, the trust-not-evaluated disclosure, and no certificate block on a `ws://` relay


src-tauri/src/lib.rs
```diff
@@ -4435,9 +4454,244 @@ static RELAY_METRICS: Lazy<RwLock<HashMap<String, RelayMetrics>>> =
...
+        return;
+    }
+    if let Ok(mut failures) = RELAY_FAILURES.write() {
+        failures.insert(normalize_relay_url(url), failure);
+    }
+}
+
+/// Clear any stored failure reason for `url`. Reconnect-to-Connected, relay removal, relay
+/// disable, and logout all route through this.
+fn clear_relay_failure(url: &str) {
```
daopunk, line 4496: `store_relay_failure_if_current` skips stale generations; this clear does not.

The first-account monitor never re-arms (`MONITOR_STARTED`) and still calls this from the `RelayStatus::Connected` arm and from successful `try_connect`. After logout → login in the same process, a shared default relay coming up on the *old* client wipes whatever reason the current account just stored.

R15 / KTD9 need the same generation check on clears (user-initiated remove/disable/logout can stay unguarded).

src-tauri/src/lib.rs
```diff
@@ -5323,6 +6265,10 @@ async fn monitor_relay_connections() -> Result<bool, String> {
     }
 
     let client = get_nostr_client().expect("Nostr client not initialized");
+    // Captured once, at spawn time, and reused by every task below. Compared against the
+    // live generation on every diagnostic write so a stale monitor loop left running for a
+    // previous account can never attribute a failure to the current one (KTD9).
```
daopunk: This comment says every diagnostic write is generation-guarded. Only `store_relay_failure_if_current` actually checks.

The tasks spawned below still call `add_relay_log` and `update_relay_metrics` with no generation test. Logout clears the snapshot; it does not stop the stale loop. Default relay URLs overlap, so the next account’s detail panel can fill with the previous client’s reconnect/health lines.

Guard those two helpers the same way as failure stores, or R15 is only half-implemented.

src-tauri/src/lib.rs
```diff
@@ -4509,9 +4763,666 @@ pub(crate) fn record_send_outcome(event: &Event, output: &Output<EventId>) {
...
+        Ok(normalized) => normalized,
+        Err(_) => return Ok(ProbeResult::failure(RelayFailureCode::InvalidUrl)),
+    };
+
+    match tokio::time::timeout(PROBE_DEADLINE, run_relay_probe(&normalized)).await {
+        Ok(Ok(round_trip_ms)) => Ok(ProbeResult::Reachable { round_trip_ms }),
+        Ok(Err(failure)) => Ok(ProbeResult::Unreachable { failure }),
+        // The single deadline -- resolution, connect, and query together --
+        // elapsed with no classifiable response at all.
+        Err(_) => Ok(ProbeResult::failure(RelayFailureCode::NotARelay)),
```
daopunk: Every outer-deadline miss becomes `not_a_relay`, including a firewalled or hung TCP/TLS connect that never reached the query.

`PROBE_INNER_TIMEOUT` is 3600s, so `try_connect_relay` / `fetch_events` almost never lose to their own timeout — the 10s wrapper does, and R5 already has `timed_out` as a distinct code. Silence after a completed handshake can stay `not_a_relay`; a deadline that fires earlier should not.

Non-blocking next to the generation-guard gap, but the label is wrong for the common “host never answers” case.

daopunk (changes requested):
## Verdict

Request changes. Capture-only TLS stays contained, and the probe never joins the live pool or writes. The login-generation guard does not hold R15 / KTD9: a stale first-account monitor can still wipe or populate diagnostics that the current account then sees.

Audited: verifier containment, probe isolation, generation guard, reconnect-loop delta, classifier/redaction, UI contract. Cross-checked with a security pass and a bug pass against this head.

## Findings

| Severity | Location | Issue |
|---|---|---|
| Medium | `src-tauri/src/lib.rs` (`clear_relay_failure`) | Clears are not generation-guarded. After logout → login in the same process, the first-account monitor still hits `Connected` / successful `try_connect` and can erase the current account’s stored reasons on shared default relays. |
| Medium | `src-tauri/src/lib.rs` (`add_relay_log` / `update_relay_metrics`) | Only failure *stores* check `LOGIN_GENERATION`. Status and health tasks keep writing logs and metrics from the stale client, so Settings → Nostr can show another account’s activity after switch. Logout clear does not stop the ongoing stream. |
| Medium | `src-tauri/src/lib.rs` (`probe_relay`) | The outer 10s deadline maps every miss to `not_a_relay`, including a hung TCP/TLS connect. R5 lists `timed_out` as its own outcome; inner timeouts are 3600s so the outer deadline usually wins before a classified transport error. |

The first two are the change request. They are one guard that was only applied to one write path. Extend the same `captured_generation != current_login_generation()` check to clears, `add_relay_log`, and `update_relay_metrics` (or wrap all four). The timeout mapping is real but non-blocking — fix with the guard or as a follow-up.

## Looked, not a problem

- **Verifier containment.** `CaptureOnlyVerifier` and `capture_only_client_config()` are module-private. Crate-visible extras (`parse_certificate`, `RelayCertificate`, `expiry_verdict`) are parse/DTO only. Source-scan and the expired-cert `RelayPool` regression keep the live path on the normal rustls stack. `local-relay-tls`, the `tokio-tungstenite` block, and `rustls = "0.23"` are unchanged.
- **Probe isolation.** Throwaway `RelayPool::new()` + `reconnect(false)`, `fetch_events` / EOSE only, `validate_relay_url` first, no `get_nostr_client()`. Explicit paths `force_remove_relay` + drop. Cancelled-future teardown via `Drop` is acceptable given `reconnect(false)` and the 10s bound.
- **`get_relay_certificate` without `validate_relay_url`.** Any `wss://` host gets a bounded handshake. Same local-operator IPC boundary as add/probe; `ws://` opens no socket. Not a change request.
- **`LOGIN_GENERATION` bump site.** Only `set_nostr_client`. `login` short-circuits when a client already exists, so a mid-session bump that would disable the still-running first-account loops is not in the current login paths.
- **Forced `disconnect()` before `try_connect`.** Only runs for relays already in `unhealthy_relays`. Needed because SDK `can_connect` is `Initialized \| Terminated \| Sleeping`. Health-check query failures on Connected stay uninstrumented (KTD1). Watch reconnect churn, but this is the capture site the plan called for.
- **`Io(InvalidData) → tls_failed`.** Broader than U1’s first mapping; matches tokio-rustls wrapping a rejected handshake as `InvalidData`. Golden vector plus the containment regression justify it.
- **Truncate-then-redact.** Cap keeps the prefix, so a `wss://` that starts inside the 200-byte window still has a scheme and the new relay-scheme path drops the whole query. `http(s)` allowlist unchanged.
- **Frontend contract.** Failure reason is label-only and omitted when status is `connected`. Probe disabled in-flight / invalid URL; add ignores probe outcome. Cert block is `wss://` only; disclosure always renders in `--text-secondary`; expiry uses `--danger` / `--warning` / `--success`. `en`/`es` key parity holds. Activity log keyed by index. Both new commands have real `invoke` callers; orphan baseline untouched.

src-tauri/src/lib.rs
logicminds: > R15 / KTD9 need the same generation check on clears (user-initiated remove/disable/logout can stay unguarded).

Addressed in 31301594 — monitor Connected / successful `try_connect` paths now call `clear_relay_failure_if_current`. Remove/disable stay unguarded. Logout also bumps `LOGIN_GENERATION` so a leftover loop cannot refill the snapshot before the next login.

src-tauri/src/lib.rs
logicminds: > Guard those two helpers the same way as failure stores, or R15 is only half-implemented.

Addressed in 31301594 — monitor tasks now call `add_relay_log_if_current` and `update_relay_metrics_if_current`, skipping writes when the captured login generation is stale. Logout also bumps the generation so a leftover loop cannot refill logs/metrics after the snapshot is cleared.

src-tauri/src/lib.rs
logicminds: > Every outer-deadline miss becomes `not_a_relay`, including a firewalled or hung TCP/TLS connect that never reached the query.

Addressed in 31301594 — a deadline that fires before the throwaway handshake completes is `timed_out`. Silence after a completed handshake stays `not_a_relay`.

logicminds:
> The first two are the change request. They are one guard that was only applied to one write path. Extend the same `captured_generation != current_login_generation()` check to clears, `add_relay_log`, and `update_relay_metrics` (or wrap all four). The timeout mapping is real but non-blocking — fix with the guard or as a follow-up.

Addressed in 31301594:

- Monitor diagnostic writes (failure clear, logs, metrics) now skip on a stale login generation. User-initiated remove/disable stay unguarded. Logout bumps `LOGIN_GENERATION` so a leftover first-account loop cannot refill the snapshot before the next login.
- Probe outer-deadline misses before handshake completion map to `timed_out`; silence after a completed handshake stays `not_a_relay`.

daopunk (approved):
## Verdict

Approve. The three change-request items from the last review are fixed on `ddb5b2c8`. Capture-only TLS stays contained, the probe never joins the live pool or writes, and monitor diagnostic writes (store, clear, log, metric) now skip on a stale login generation. Logout also bumps `LOGIN_GENERATION`, so a leftover first-account loop cannot refill the snapshot.

Re-audited HEAD: verifier containment, probe isolation, generation-guard completeness, probe timeout mapping, reconnect-loop delta, classifier/redaction, UI contract. Cross-checked with a security pass against this head.

## Prior findings

| Prior | HEAD status |
|---|---|
| Medium — `clear_relay_failure` unguarded | Fixed. Monitor Connected / successful `try_connect` paths call `clear_relay_failure_if_current`. User-initiated remove/disable/logout stay unguarded. Covered by `logout_bumps_generation_so_pre_logout_monitor_writes_are_skipped`. |
| Medium — `add_relay_log` / `update_relay_metrics` unguarded | Fixed. Monitor tasks call `add_relay_log_if_current` and `update_relay_metrics_if_current`. Logout bump prevents a leftover loop from refilling logs/metrics after the snapshot clear. |
| Medium — probe deadline → `not_a_relay` | Fixed. `handshake_done` is set only after `try_connect_relay` succeeds. Deadline miss before handshake → `timed_out`; silence after handshake → `not_a_relay`. `accept_then_stall_host` now asserts `timed_out`. |

## Security

No medium, high, or critical issues in the relay-diagnostics paths.

| Area | Result |
|---|---|
| Verifier containment | Pass. `CaptureOnlyVerifier` and `capture_only_client_config()` are module-private. Source-scan plus the expired-cert `RelayPool` regression keep live traffic on the normal rustls stack. `local-relay-tls`, the `tokio-tungstenite` block, and `rustls = "0.23"` are unchanged. |
| Probe isolation | Pass. Throwaway `RelayPool::new()` + `reconnect(false)`, `fetch_events` / EOSE only, `validate_relay_url` first, no `get_nostr_client()`. Explicit teardown on success/error; cancelled-future `Drop` is acceptable given `reconnect(false)` and the 10s bound. |
| Generation guard | Pass for the new/changed monitor diagnostic writes. `set_nostr_client` and `clear_nostr_client` both bump `LOGIN_GENERATION`. |
| Classifier / redaction | Pass. Typed downcast, no message-text codes. `ws`/`wss` drop the whole query and strip userinfo. Truncate-then-redact still keeps a scheme in the 200-byte window. UI renders `code`, not `detail`. |
| Account isolation / TLS trust | Pass. Read-side gate omits reasons when status is `connected`. Certificate panel always shows the trust-not-evaluated disclosure. Capture path does not widen live-pool trust. |

## Looked, not a problem

- **Monitor loops do not re-arm.** `MONITOR_STARTED` is process-lifetime by design (plan stop condition / known limitation). R1 already discloses that classified reasons from the reconnect loop are first-account-in-process only. Generation skip + logout clear is the intended R15 mechanism; do not treat re-arming as leftover work from this PR.
- **`record_event_received` / `record_send_outcome` still unguarded.** Predates this feature. Monitor stale-loop risk for failure reasons/logs/metrics is what KTD9 addresses.
- **`get_relay_certificate` without `validate_relay_url`.** `wss_host_port` rejects non-`wss` before any socket. Same local-operator IPC boundary as add/probe. Not a change request.
- **Forced `disconnect()` before `try_connect`.** Only for statuses the SDK `can_connect` set cannot act on. Needed so failures are actually attempted and classified.
- **Frontend contract.** Failure reason is label-only and omitted when status is `connected`. Probe disabled in-flight / invalid URL; add ignores probe outcome. Cert block is `wss://` only; disclosure always renders in `--text-secondary`; expiry uses `--danger` / `--warning` / `--success`. `en`/`es` key parity holds. Activity log keyed by index. Both new commands have real `invoke` callers.

## Optional follow-ups (not blocking)

- Call `validate_relay_url` in `get_relay_certificate` for userinfo rejection parity with probe/add.
- Redact relay URLs in health-check log lines (`Health check failed: {e}`) if errors ever echo a configured URL with query auth.
- Generation-guard `record_event_received` if full process-lifetime isolation beyond monitor paths is wanted later.

PR #275, opened 2026-08-17
fix: stop 0.5.x upgrades from tripping the storage-format gate

logicminds:
## Summary

Users upgrading from 0.5.x to 0.6.0 (fresh install or in-app update + restart) hit "This build cannot open your data" before they could even unlock their account — a hard, machine-wide block claiming their data was "written by a newer version of Pacto." It wasn't. Enabling refinery's `int8-versions` Cargo feature (`SchemaVersion` `i32` -> `i64`, needed so a 14-digit UTC-timestamp migration version fits) silently changed the checksum of every migration a pre-0.6.0 build had already applied — refinery hashes `(name, version, sql)` together, and hashing an `i32` vs an `i64` for the identical version value feeds a different byte width into `SipHasher13`, producing a different digest with no change to the migration itself. Every account that had applied any migration under a pre-0.6.0 build (i.e. nearly everyone) hit a checksum mismatch on `V1` at cold launch.

`storage_format::legacy_i32_checksum` reproduces the pre-`int8-versions` digest so `classify_history` recognizes a row stamped with either checksum, fixing the pre-auth gate. `migrations::reconcile_legacy_checksums_for_table` runs a raw-SQL repair before refinery's own `Runner::run`, so its internal verification — which enforces only the current checksum — also accepts the legacy history, and rewrites it to the current checksum so the repair only ever runs once per account. A row matching neither checksum is still a genuine divergence and still gates.

**Second commit, from code review on this branch:** the same break also invalidates `_refinery_schema_history_nostr_mls`, the separate refinery history table `mdk-sqlite-storage` 0.8.0 owns on the MLS store (`mls.db`) — Cargo unifies features per dependency across the whole build, so `int8-versions` silently applies there too, and that crate's own `run_migrations` aborts independently of `pacto.db`'s gate. `mdk-sqlite-storage` 0.8.0 predates 0.6.0 (shipped unchanged in 0.5.4/0.5.5), so this hit the exact same upgrade population — and the first commit alone made it *harder* to notice, since it let affected accounts past the pre-auth gate and into a successful login, only to then find their MLS engine permanently unopenable (the store's own `Current`/`Legacy` classification looks only at the history table's version number, never its checksum). `mls_legacy_checksum` vendors a byte-for-byte copy of `mdk-sqlite-storage`'s five migration files (that crate doesn't expose its embedded set publicly) to reconcile the same way, called right before `MdkSqliteStorage::new_with_key`.

## New concepts

**A Cargo feature flag can silently invalidate a stored hash/checksum by changing an integer's type width.**

`refinery-core`'s migration checksum hashes `(name, version, sql)` through `SipHasher13`. Rust's integer `Hash` impls write the value's raw bytes (`write_i32` vs `write_i64`), and `SipHasher13` mixes the total byte length into its finalization — so hashing the *same numeric value* as a different integer type produces a *different* digest, even though nothing about the logical data changed. `refinery`'s `int8-versions` feature retypes `SchemaVersion` from `i32` to `i64`; every migration checksum this build computes shifted the moment that feature turned on, silently breaking equality against every checksum a build without the feature had already written to disk. Cargo's feature unification means this doesn't stay contained to the crate that requests the feature — every dependency sharing that same `refinery-core` resolution gets it too, which is exactly how a change to *this* app's Cargo.toml broke a vendored third-party crate's own migration checksums.

This class of bug is worth watching for anywhere a checksum, hash, or cache key is derived from a value whose declared type can change (a feature flag, a generic parameter, a newtype swap) — the stored artifact and the freshly computed one diverge with no compiler error and no obvious symptom besides "recognized data now looks unrecognized."

Verified independently of this repo's code: two throwaway crates pinned to the exact `refinery-core = 0.9.2`, one with `int8-versions` and one without, hashing the identical `(name, sql)` — different checksums (`470904781690771365` vs `17539547413937229480`). That measured pair is now pinned as a golden-value regression test so the in-repo reproduction can't silently drift from the real algorithm.

Don't reach for this fix pattern (tolerate both digests + reconcile) unless the checksum's *only* job is drift detection, not a security boundary — silently accepting a legacy digest is safe here because a real divergence (wrong name, wrong SQL) still fails both checks.

## Validation

- `v1_database_with_legacy_checksum_reconciles_and_migrates` (pacto.db): stamps a `V1` row with the legacy checksum, confirms `run_migrations` succeeds and reconciles it.
- `mls_store_with_legacy_checksums_fails_to_reopen_without_reconciliation` / `reconcile_mls_store_legacy_checksums_lets_mdk_reopen` (MLS store): builds a *real* store via `mdk_sqlite_storage::MdkSqliteStorage::new_with_key`, stamps it with legacy checksums, proves the real MDK crate refuses to reopen it without the fix and succeeds with it — not just this repo's model of refinery's behavior.
- `legacy_i32_checksum_matches_independently_measured_golden_value`: pins the reproduction to the externally-measured digest pair above.
- `mdk_sqlite_storage_version_matches_vendored_migrations`: fails the build if `Cargo.lock`'s pinned `mdk-sqlite-storage` version ever drifts from the vendored migration copy.
- `cargo test --lib`: 748 passed, 0 failed, 2 ignored (pre-existing).
- `cargo fmt --check` clean on every touched file.

## Follow-up (not in this PR)

This fixes the checksum going forward; every account already stuck on the current 0.6.0 binary needs a build with this fix (0.6.1) to self-heal on next launch — there's no way for an already-bricked install to reconcile itself without one. Worth flagging in the release notes and to anyone already blocked.

### DIFF
--- a/src-tauri/src/mls_store_reset.rs
+++ b/src-tauri/src/mls_store_reset.rs
@@ -95,6 +95,51 @@ fn classify_store(store_path: &Path, encryption_key: &[u8; 32]) -> StoreClassifi
     StoreClassification::Legacy
 }
 
+/// Reconciles `_refinery_schema_history_nostr_mls` against
+/// `mls_legacy_checksum`'s vendored `mdk-sqlite-storage` migration set,
+/// before `MdkSqliteStorage::new_with_key` gets a chance to run its own
+/// checksum-verifying refinery `Runner` over that table. Called only from
+/// `reset_with_connection`'s `Current` arm, after `classify_store` has
+/// already confirmed the history table is readable at `store_path`.
+///
+/// Tries a plain (unkeyed) connection first, then falls back to
+/// `encryption_key` -- mirroring `classify_store`'s own plain-then-keyed
+/// order. A real MDK 0.8.0+ store is always SQLCipher-encrypted, so the
+/// plain attempt's `UPDATE`s fail closed with a "file is not a database"
+/// error (SQLCipher rejects an unkeyed read past the header) and the keyed
+/// attempt is what actually reconciles it; a plain attempt succeeding is
+/// reserved for test fixtures that model a `Current` store without real
+/// encryption. See `mls_legacy_checksum` for why this repair exists at all.
+fn reconcile_mls_store_legacy_checksums(
+    store_path: &Path,
+    encryption_key: &[u8; 32],
+) -> Result<(), String> {
+    let migrations = crate::mls_legacy_checksum::embedded_migration_set();
+
+    if let Ok(conn) = Connection::open(store_path) {
+        if crate::migrations::reconcile_legacy_checksums_for_table(
+            &conn,
+            crate::mls_legacy_checksum::MLS_HISTORY_TABLE_NAME,
+            &migrations,
+        )
+        .is_ok()
+        {
+            return Ok(());
+        }
+    }
+
+    let conn = Connection::open(store_path)
+        .map_err(|e| format!("Failed to open MLS store for checksum reconciliation: {e}"))?;
+    let key_hex = hex::encode(encryption_key);
+    conn.execute_batch(&format!("PRAGMA key = \"x'{key_hex}'\";"))
+        .map_err(|e| format!("Failed to key MLS store for checksum reconciliation: {e}"))?;
+    crate::migrations::reconcile_legacy_checksums_for_table(
+        &conn,
+        crate::mls_legacy_checksum::MLS_HISTORY_TABLE_NAME,
+        &migrations,
+    )
+}
+
 fn archive_path(profile_dir: &Path, now: u64) -> PathBuf {
     let base = profile_dir.join(format!("{ARCHIVE_PREFIX}{now}"));
     if !base.exists() {
@@ -380,7 +425,15 @@ fn reset_with_connection(
                 "Unsupported MLS store schema version {version}; refusing to open or archive. Expected 1–5 (current) or ≥100 (legacy)."
             ));
         }
-        StoreClassification::Fresh | StoreClassification::Current => {
+        StoreClassification::Fresh => {
+            put_setting(conn, RESET_MARKER_KEY, "complete")?;
+            if let Err(e) = prune_archives(profile_dir, now) {
+                eprintln!("[MLS Reset] Failed to prune archives: {e}");
+            }
+            return Ok(ResetOutcome::default());
+        }
+        StoreClassification::Current => {
+            reconcile_mls_store_legacy_checksums(store_path, encryption_key)?;
             put_setting(conn, RESET_MARKER_KEY, "complete")?;
             if let Err(e) = prune_archives(profile_dir, now) {
                 eprintln!("[MLS Reset] Failed to prune archives: {e}");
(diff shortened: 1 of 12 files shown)


logicminds:
Ran an independent code review against this branch. Summary of findings and how each was handled:

**1. (Critical, confidence 0.82) Same checksum-width break also affects the MLS store's `_refinery_schema_history_nostr_mls`, not just `pacto.db`.** Confirmed independently (`cargo tree -e features -i refinery-core` shows Cargo unifies `int8-versions` into the same `refinery-core` `mdk-sqlite-storage` depends on; `mdk-sqlite-storage` 0.8.0 was pinned unchanged in 0.5.4/0.5.5, predating the feature). Fixed in `b1bc00e8`: vendored `mdk-sqlite-storage`'s migration set (`mls_legacy_checksum.rs` + `src/mls_migrations/`), generalized the reconciliation helper, and wired it into `mls_store_reset` before `MdkSqliteStorage::new_with_key` runs. Verified against the *real* `mdk_sqlite_storage` crate, not just this repo's model of it (see `mls_store_with_legacy_checksums_fails_to_reopen_without_reconciliation` / `reconcile_mls_store_legacy_checksums_lets_mdk_reopen`).

**2. Tests validated the reproduction against itself, not an external anchor.** Fixed: added `legacy_i32_checksum_matches_independently_measured_golden_value`, pinning the digest pair measured via two throwaway crates built directly against the pinned `refinery-core = 0.9.2` (one with `int8-versions`, one without) — not derived from the code under test.

**3. Reconciliation loop re-runs on every connection open, not truly "one-time" as documented.** Behaviorally harmless (idempotent, indexed no-op updates, no partial-write/atomicity risk) per the review's own assessment — fixed the misleading doc comment, left the loop as-is; not worth a caching mechanism for ~34 no-op statements on a pooled-connection miss.

Full review transcript: history://JoyousBat

src-tauri/src/mls_store_reset.rs
```diff
@@ -380,7 +425,15 @@ fn reset_with_connection(
...
-        StoreClassification::Fresh | StoreClassification::Current => {
+        StoreClassification::Fresh => {
+            put_setting(conn, RESET_MARKER_KEY, "complete")?;
+            if let Err(e) = prune_archives(profile_dir, now) {
+                eprintln!("[MLS Reset] Failed to prune archives: {e}");
+            }
+            return Ok(ResetOutcome::default());
+        }
+        StoreClassification::Current => {
+            reconcile_mls_store_legacy_checksums(store_path, encryption_key)?;
```
daopunk, line 455: **Blocking.** This is the only call site, and `reset_with_connection` returns earlier when `settings.mls_store_reset_v1 = complete` (pacto.db). That marker shipped in v0.5.5 (6df3f0e7). Every 0.5.x account that already opened MLS has it set, so they never reach this arm.

Result: pacto.db gate is fixed, then `MdkSqliteStorage::new_with_key` still dies with `DivergentVersion`.

Move the call to `ensure_store_ready` (after `reset_with_connection`) or to immediately before `new_with_key` in `mls.rs`. Add a test that sets the marker, stamps legacy checksums, drives `reset_with_connection` / `ensure_store_ready`, and asserts the real MDK crate reopens. The new tests call `reconcile_mls_store_legacy_checksums` directly and miss this.

src-tauri/src/mls_store_reset.rs
```diff
@@ -95,6 +95,51 @@ fn classify_store(store_path: &Path, encryption_key: &[u8; 32]) -> StoreClassifi
...
+) -> Result<(), String> {
+    let migrations = crate::mls_legacy_checksum::embedded_migration_set();
+
+    if let Ok(conn) = Connection::open(store_path) {
+        if crate::migrations::reconcile_legacy_checksums_for_table(
+            &conn,
+            crate::mls_legacy_checksum::MLS_HISTORY_TABLE_NAME,
+            &migrations,
+        )
+        .is_ok()
```
daopunk: `Connection::open` is read-write, unlike `classify_store`'s `READ_ONLY` probe. `reconcile_legacy_checksums_for_table` returns `Ok` for 0-row UPDATEs, so a connection that can execute SQL but cannot see the real history table would skip the keyed path.

SQLCipher should fail the unkeyed UPDATE and fall through, but I would still: (1) `SELECT` the history table before treating this connection as authoritative, (2) only write on the connection that read real rows.

daopunk (changes requested):
## Verdict

The diagnosis is correct, and the `pacto.db` half looks sound: `legacy_i32_checksum` + `classify_history` leniency + `reconcile_legacy_checksums_for_table` before `Runner::run` is the right shape, the golden-value test is a real external anchor, and leaving a neither-current-nor-legacy row alone still fails closed.

The MLS half does **not** run for the population this PR says it is fixing. I would not merge until that wiring is moved off the one-shot reset path.

## Blocking: MLS reconcile is skipped when `mls_store_reset_v1=complete`

`reconcile_mls_store_legacy_checksums` is only called from `reset_with_connection`'s `Current` arm. That arm is after:

```rust
if setting(conn, RESET_MARKER_KEY)?.as_deref() == Some("complete") {
    return Ok(ResetOutcome::default());
}
```

`conn` is **pacto.db** (`account_manager::get_db_connection`). `mls_store_reset_v1` has shipped since 6df3f0e7 and is **in v0.5.5**. Any 0.5.x account that opened MLS after that commit already has the marker set — including the exact 0.5.4/0.5.5 + mdk 0.8.0 upgrade path called out in the PR.

Those accounts will:
1. pass the `pacto.db` pre-auth gate (first commit works),
2. skip MLS reconcile on every launch,
3. hit `MdkSqliteStorage::new_with_key` → `DivergentVersion`.

The new MLS tests call `reconcile_mls_store_legacy_checksums` **directly**. They never go through `reset_with_connection` / `ensure_store_ready`, so they cannot catch this.

**Fix:** run the repair on every open of a `Current` store, not once per reset generation. Cleanest: call it from `ensure_store_ready` after `reset_with_connection` returns (or immediately before `MdkSqliteStorage::new_with_key` in `mls.rs`). Add a test that sets `mls_store_reset_v1=complete`, stamps legacy checksums, drives `reset_with_connection`, then asserts the real MDK crate reopens.

Until that lands, 0.6.1 still leaves upgraded MLS stores unopenable. The follow-up note in the PR body understates this — it is not just “need a build with the fix”; the fix as wired will not self-heal the stated users.

## Other notes (non-blocking if the above is fixed)

**Unkeyed write probe.** `reconcile_mls_store_legacy_checksums` uses `Connection::open` (read-write) and treats any `Ok` from the UPDATE loop as “done,” including 0-row no-ops. `classify_store` probes `READ_ONLY` first. Prefer: confirm `SELECT` on `_refinery_schema_history_nostr_mls` succeeds before writing; only write on the connection that can actually see the table. Production SQLCipher should fail the unkeyed UPDATE and fall through, so this is belt-and-suspenders, not the merge blocker.

**`pacto.db` test coverage.** `v1_database_with_legacy_checksum_reconciles_and_migrates` only stamps V1. A real 0.5.x history has every applied row on the legacy digest. The loop should handle that; a fixture with several stamped versions would prove it.

**Vendored MLS SQL.** `reconcile_mls_store_legacy_checksums_lets_mdk_reopen` is the real content check (crate reopen fails if vendored SQL ≠ crate SQL). The Cargo.lock `contains("0.8.0")` canary is fine as a bump alarm, not as proof of byte-identity.

**Checksum is not a security boundary.** Agree. Accepting the second digest of the same `(name, version, sql)` does not weaken the gate against a real name/SQL divergence.

## What looks good

- Root cause (SipHash byte-width via `int8-versions` + Cargo feature unification) is accurately explained and independently reproduced.
- `pacto.db` path has no equivalent early-return; reconcile runs on the same connection `Runner::run` will use.
- Fresh vs Current split is correct — a missing MLS store should not try to UPDATE a history table.
- Genuine neither-checksum rows still classify as `Divergent`.
- PR description is value-first and honest about the 0.6.0-already-stuck population needing 0.6.1.

src-tauri/src/mls_store_reset.rs
logicminds: > Move the call to `ensure_store_ready` (after `reset_with_connection`) or to immediately before `new_with_key` in `mls.rs`. Add a test that sets the marker, stamps legacy checksums, drives `reset_with_connection` / `ensure_store_ready`, and asserts the real MDK crate reopens.

Addressed in 53925165 — moved the reconciliation into `reset_with_connection`'s marker-complete fast path itself, so it now runs on every open of a `Current` store (guarded by `store_path.exists()`, since a Fresh account hitting this path hasn't had its store created yet) instead of only the one-shot `Current`-classification branch. Added `reset_with_connection_reconciles_current_store_on_marker_complete_fast_path`, which sets `mls_store_reset_v1=complete`, stamps legacy checksums on a real SQLCipher store, drives `reset_with_connection`, and asserts the real `mdk_sqlite_storage` crate reopens it.

src-tauri/src/mls_store_reset.rs
logicminds: > I would still: (1) `SELECT` the history table before treating this connection as authoritative, (2) only write on the connection that read real rows.

Addressed in 53925165 — `reconcile_mls_store_legacy_checksums` now runs `history_table_readable` (a `SELECT COUNT(*)` on the history table) on the unkeyed connection first, and only uses that connection to write if the read actually succeeded; otherwise it falls through to the keyed connection unconditionally, same as before.

logicminds:
Re your review — the two blocking items are addressed and replied to on their threads (marker-complete fast path now reconciles; unkeyed connection now SELECTs before writing), pushed in 53925165.

On the non-blocking notes:

> `v1_database_with_legacy_checksum_reconciles_and_migrates` only stamps V1. A real 0.5.x history has every applied row on the legacy digest. The loop should handle that; a fixture with several stamped versions would prove it.

Added `reconcile_legacy_checksums_for_table_rewrites_every_stamped_row`, a direct unit test that stamps five real embedded migrations with legacy checksums plus one genuinely divergent row, and asserts the loop rewrites all five while leaving the divergent row untouched.

The vendored-SQL and checksum-is-not-a-security-boundary notes are agreement, not asks — no change needed there.

daopunk (approved):
## Verdict

Previous blockers are fixed in 53925165. The 0.5.x → 0.6.1 path that was going to skip MLS repair now actually runs it. Approving.

## Prior review items

**Blocking: MLS reconcile skipped when `mls_store_reset_v1=complete`.** Fixed. The marker-complete fast path now reconciles whenever `store_path.exists()`, so every 0.5.x account that already opened MLS still gets the repair instead of dying on `MdkSqliteStorage::new_with_key`. `reset_with_connection_reconciles_current_store_on_marker_complete_fast_path` covers this: marker set, legacy checksums stamped on a real SQLCipher store, driven through `reset_with_connection`, then reopened via the real `mdk_sqlite_storage` crate.

**Unkeyed write probe.** Fixed. `history_table_readable` (`SELECT COUNT(*)`) gates the unkeyed connection; only that connection writes, otherwise keyed fallback.

**Multi-row `pacto.db` fixture.** Fixed. `reconcile_legacy_checksums_for_table_rewrites_every_stamped_row` stamps five real migrations plus one genuinely divergent row and asserts the loop rewrites all five while leaving the divergent row alone.

Production order is still `ensure_store_ready` → `MdkSqliteStorage::new_with_key` in `mls.rs`, so the repair runs before MDK’s own runner. `pacto.db` path was already sound.

## Remaining nits (non-blocking)

- Keyed fallback still writes without a SELECT. Unkeyed is gated; keyed is not. Wrong-key SQLCipher should fail the `UPDATE` rather than no-op, so this is belt-and-suspenders.
- Stale module comment in `mls_legacy_checksum.rs`: it still says `ensure_store_ready` classifies `Current` and “does nothing further.” That was the bug. The `reset_with_connection` comment is now accurate.
- Fast path reconciles any existing file, not only a classified `Current` store. For the 0.5.x marker-complete population that is the right store. A leftover non-history file would now fail `reset_with_connection` instead of falling through to MDK; fail-closed, fine.

No new merge blocker.

PR #278, opened 2026-08-18
feat: squad pfp and pfp picker

### LINKED ISSUE #209: Add profile pictures (PFPs) for squads
## Problem

User profiles can set a PFP (image upload via Blossom → `upload_avatar` → `update_profile`). Squads do not have an equivalent flow.

Today squads only support an optional **icon URL** string at create/organize time (`icon_url` / `iconUrl` on the squad catalog and commons wire). That is paste-a-URL, not a first-class avatar: no image picker, no upload, no edit-after-create parity with user profile photos. Most squad surfaces therefore fall back to letter placeholders or empty images.

## Why it matters

Squads are first-class social objects in the navbar, commons, pairing, and hub UI. Without PFPs they look unfinished next to user avatars and are harder to scan in lists.

## Proposed direction

Give squads the same class of avatar UX users already have:

1. **Upload / set a squad PFP** — pick an image, upload (reuse Blossom/`upload_avatar` or a squad-scoped upload type), persist as the squad’s avatar URL (`icon_url` or a clearer field if renamed).
2. **Edit / remove** after create — not only at organize time.
3. **Show consistently** wherever squads are listed or headed (navbar, hub, commons cards, pairing, invites, etc.), with a letter/initial fallback when unset.
4. **Sync** so members see the same image (MLS / commons / catalog announcement path — align with how `icon_url` already travels on create broadcasts).

Exact product choices (who can change it — admins only vs any member; max size; square crop) can follow user-profile rules unless squad policy needs stricter admin gating.

## Acceptance criteria

- [ ] User can set a squad PFP via image upload (not URL-paste-only)
- [ ] User can change or clear the squad PFP after the squad exists
- [ ] Squad PFP appears in primary squad UI surfaces with a sensible fallback when missing
- [ ] Other members receive/see the updated PFP through existing squad metadata sync
- [ ] i18n for new copy (en + es); icon-only Edit control if an edit affordance is added (`EditIconButton`)
- [ ] Single-account / no-icon squads remain fine; no forced upload at create

## Notes / existing hooks

- Schema/API already has `icon_url` on squad catalog (`src-tauri/src/squad_catalog.rs`) and commons create payloads (`squad_icon_url`).
- Create UI currently exposes a raw URL field (`nav.navbar.organizeSquad.iconLabel` / `squad.pair.iconLabel`).
- User path to mirror: `upload_avatar` + profile update in `src/lib/api/nostr.ts` and settings profile UI.

### LINKED ISSUE #276: Move squad-dashboard Stickers into a Settings tab
## Summary

Replace the `#squad-dashboard` **Stickers** tab with a **Settings** tab. Status stays the frequently used operational surface (Checklist, Broadcasts). Occasional config — Chain RPC, Pimlico key, Join-inbox holders, Contracts, stickers, plus the network editor and squad avatar already on Status — moves into Settings as sections.

This also documents an invariant: do not add a new dashboard segmented mode for a single feature. Occasional config belongs in Settings.

Composer Stickers in `MessageInput.svelte` is unchanged. Broadcasts stay on Status.

## Split

**Status** — inform the squad: Checklist progress, Broadcast visibility, and other frequently needed operational info. Checklist CTAs that configure something (select network) switch to Settings. Deploy / mint-hats CTAs stay as they are (launchpad / Governance).

**Settings** — one-time or occasional config. New features of this kind become **sections here**, not new segmented modes.

**Tab order:** Status → Governance → Treasury → Roles → Crew → **Settings** (replaces Stickers).

### 1. Mode slug (greenfield, no alias)

In `src/stores/navigation.ts`:

- Replace `'stickers'` with `'settings'` on `SquadDashboardChannelMode` and `parseSquadDashboardChannelMode`.
- Persisted `stickers` (and any other unknown value) **resets to `status`**. Do not map `stickers` → `settings`.
- Rename `focusSquadStatusRpcEditor` / `squadStatusRpcFocusNonce` → Settings (`focusSquadSettingsRpcEditor`). `RpcReadErrorCard` and `SquadEndpointsPanel` follow.

Update `src/stores/navigation.test.ts` and `src/stores/parent-dashboard-mode.test.ts` (today `settings` is treated as unknown and reset to `status`).

### 2. New Settings tab, slim Status

- Add runes `DashboardSettingsTab.svelte`. Host, in order: identity, network editor, `SquadEndpointsPanel`, `SquadBotHoldersSection`, `SmartContractSecuritySection`, stickers section.
- Rename `DashboardStickersTab.svelte` → `SquadStickersSection.svelte` (keep runes). Adjust heading/aria from “tab” to “section”.
- Slim `DashboardStatusTab.svelte` to Checklist + Broadcasts. Convert it to runes as part of that rewrite. “Select network” checklist action calls `selectDashboardView('settings')`.
- `ParentDashboard.svelte`: swap `stickers` for `settings` in `DASHBOARD_VIEWS`, lazy loader, keep-alive pane. Extend member/EVM loaders that today key off `status || crew` so Join-inbox still hydrates on Settings. RPC save handlers that Status currently receives move onto the Settings pane.
- `dashboard-tab-components.ts`: `loadDashboardSettingsTab`; drop `loadDashboardStickersTab`.

Do not convert the ParentDashboard god shell.

### 3. i18n

`en/governance.json` + `es/governance.json`:

- Add `governance.dashboardView.settings` / `governance.tabLoadError.settings`.
- Delete unused `…stickers` dashboard-view keys.
- Tweak `squad.stickers.tabAria` / `tabLabel` to section copy. Composer keys under `messaging.messageInput.*` stay (picker tab).

### 4. Docs — Status vs Settings invariant

Normative home: `docs/shell/LAYOUT.md`.

- Tab list: Status → Governance → Treasury → Roles → Crew → Settings.
- **Status:** Checklist, Broadcasts (visibility), frequently needed operational info.
- **Settings:** occasional/one-time config (RPC, Pimlico, Join-inbox key, contracts, stickers, network, identity).
- **Invariant:** do not add a new `#squad-dashboard` segmented mode for a single feature. Occasional config → Settings section. A new tab is only for a frequently used operational domain with its own data (Governance, Treasury, …). Stickers-as-its-own-tab is the anti-pattern.

Also update:

- `docs/messaging/STICKER_PACKS.md` §6: authoring is Settings → Stickers section, not a dashboard Stickers tab.
- `docs/communities/SQUAD_CATALOG.md`: avatar path is Dashboard → Settings (not Status).
- `docs/wallet/CHAIN_CONFIG.md` already says Settings for network; leave that, and make LAYOUT the cross-link.

Leave historical plans/PR reviews alone.

### 5. Validation

- `pnpm test` for navigation/mode parsers.
- `pnpm check` / `pnpm lint` on touched Svelte + i18n.
- If Tauri MCP is available: `#squad-dashboard` shows Settings not Stickers; Status is Checklist + Broadcasts; Settings has RPC / Pimlico / Join-inbox / Contracts / Stickers; RPC error card lands on Settings.

daopunk:
## Summary

Squads can now have a profile photo using the same crop-and-upload picker as user avatars, and that photo actually reaches other members. `#squad-dashboard` also stops growing a tab per feature: **Stickers** is gone, **Settings** is the home for occasional config, and **Status** stays the frequently used operational surface.

Before this change, Organize Squad / Pair accepted an optional URL paste into `iconUrl`. That field never left the local catalog: invites omitted it, MLS had no identity announce, and there was no way to change or clear the photo after create. Commons tiles and the hub heading mostly fell back to a letter. Dashboard Stickers was its own segmented mode, and Status mixed Checklist/Broadcasts with one-time config (RPC, Pimlico, Join-inbox, Contracts, network).

After this change, you pick a local image at create (optional) or on Dashboard → Status (edit/clear, persist immediately). Crop/upload is the shared 512×512 JPEG Blossom path. Members last-write-wins via `squad_identity_updated` on `#announcements`; new invitees get `iconUrl` on `squad_invite`. Any current MLS member can set it (same policy as sticker packs). Commons discovery tiles still snapshot `squadIconUrl` at publish time.

Dashboard tabs are now Status → Governance → Treasury → Roles → Crew → **Settings**. Status keeps squad photo, Checklist, and Broadcasts. Settings is Stickers → Network → Chain RPC / Pimlico → Join-inbox → Contracts, each in the same elevated card. New occasional config belongs in a Settings section, not a new tab. Persisted `stickers` resets to `status`. RPC-error “Add custom RPC” opens Settings. Rail squad photos clip to the circular button.

Closes #209
Closes #276

## Changes

- Extract runes `AvatarPicker` + `AvatarCropModal` and wire Profile settings through them (no URL-paste path).
- Use the picker on Organize Squad, Pair with Squad, and a new Status identity block.
- Show `SquadAvatar` on the hub heading, Commons cards/detail, invite cards, and rail fallback.
- Sync `iconUrl` with MLS `squad_identity_updated`, invite payload, accept-invite persist, and late-joiner state-sync republish.
- Replace the Stickers dashboard mode with Settings; slim Status to photo + Checklist + Broadcasts.
- Clip left-rail `Tab` images in a circular overflow wrap (unread dots on DM tabs stay unclipped).

## Test plan

- [ ] `pnpm test` and `pnpm lint`
- [ ] Organize a squad with a photo; confirm rail + hub heading (circular, not a square overflow)
- [ ] Dashboard → Status: replace and clear the photo; confirm other members update
- [ ] Invite a member; invite card shows the squad PFP and accept persists it
- [ ] Profile → Change avatar still opens the shared crop modal (e2e `avatar-crop.spec.ts`)
- [ ] `#squad-dashboard` shows Settings, not Stickers; Status has photo + Checklist + Broadcasts
- [ ] Settings order is Stickers → Network → RPC / Pimlico → Join-inbox → Contracts, each in a card
- [ ] Checklist “Select network” and RPC-error “Add custom RPC” land on Settings

---


daopunk:
Screenshot 2026-08-17 at 8 52 13 PM
Screenshot 2026-08-17 at 8 52 04 PM

daopunk:
Screenshot 2026-08-17 at 8 58 50 PM
Screenshot 2026-08-17 at 8 58 40 PM

src/components/ui/AvatarPicker.svelte
```diff
@@ -0,0 +1,207 @@
...
+  {/if}
+</div>
+
+<AvatarCropModal
+  open={cropOpen}
+  filepath={cropFilepath}
+  title={cropTitle}
+  onConfirm={handleCropConfirm}
+  onCancel={handleCropCancel}
+/>
```
logicminds, line 131: **[P1 blocking]** Nested `Modal` + Escape key discards form state.

`AvatarPicker` mounts `AvatarCropModal` (a full `ui/Modal`) inline. Wherever this component is used inside another `Modal` (`Navbar.svelte:717-731` Organize Squad, `PairWithSquadModal.svelte:88-97`), the crop modal's DOM is a descendant of the outer modal's `.modal-content`. `Modal.svelte`'s `handleContentKeydown` never calls `stopPropagation()` on Escape, so one Escape press fires the crop modal's `onCancel` **and** the outer modal's `onClose` — the entire enclosing form (squad name, member selection, tags, network) is discarded. It also defeats `dismissible={!uploading}`: the outer modal still closes mid-upload, unmounting the picker while a Blossom upload is in flight.

Fix once, in `Modal.svelte`'s `handleContentKeydown`: `stopPropagation()` on a handled Escape (or portal the crop modal to `document.body`).

src/components/ui/AvatarPicker.svelte
```diff
@@ -0,0 +1,207 @@
...
+    fallbackText?: string;
+  } = $props();
+
+  let cropOpen = $state(false);
+  let cropFilepath = $state('');
+  let pickError = $state<string | null>(null);
+  let imageBroken = $state(false);
+
+  const displaySrc = $derived(src?.trim() && !imageBroken ? src.trim() : null);
+  const showClear = $derived(editable && allowClear && !!displaySrc && !disabled);
```
logicminds: **[P2]** `showClear` gates on `displaySrc`, which nulls itself the moment `` sets `imageBroken = true`. So whenever the stored URL is unreachable (404'd Blossom/CDN icon, offline, expired host), the "Remove photo" button disappears — the user has no way to clear a dead `iconUrl`. Reachable via `SquadIdentitySection.svelte:40-55`, which passes `allowClear={!!squad.iconUrl}` expecting the control regardless of load state.

Gate on the raw prop instead:
```suggestion
  const showClear = $derived(editable && allowClear && !!src?.trim());
```
Related: `showClear` also folds in `&& !disabled`, making the separate `disabled={disabled}` binding on the button dead — it's removed rather than disabled, causing layout shift during saves.

src/lib/squad/squad-state-sync.ts
```diff
@@ -332,10 +337,16 @@ export async function respondToSquadStateSyncRequest(
...
   }
 
+  if (parent) {
+    try {
+      const ok = await publishSquadIdentityUpdated(parent);
+      if (ok) anyOk = true;
+    } catch (e) {
+      console.warn('[squad-state-sync] identity republish failed', e);
+    }
+  }
```
logicminds: **[P1 blocking]** Sync-response identity republish can wipe the squad photo squad-wide.

`respondToSquadStateSyncRequest` always republishes `icon_url: parent.iconUrl ?? null` (via `publishSquadIdentityUpdated`, `squad-identity-announce.ts:61-64`) on every sync request. A member whose local squad row has no `iconUrl` (joined before the photo was set, invite carried none, missed the announce while offline) therefore broadcasts `icon_url: null` to `#announcements`, and every receiver's `applySquadIdentityUpdated` applies it as a clear — wiping the squad photo for the whole squad. Trigger is routine: any new joiner's auto sync request fans out to all online peers.

Sibling republishers deliberately avoid this: `publishSquadNetworkUpdated` returns `false` when no chain resolves (`squad-network-share.ts:90`), `publishSquadRpcUpdated` returns `false` when no config (`squad-rpc-share.ts:93`). Identity is the only destructive republish. Fix: skip the identity republish (or the whole call) when the local `iconUrl` is empty.

src/components/parent/ParentDashboard.svelte
```diff
@@ -727,26 +727,26 @@ import { TREASURY_SAFE_UI_CAP, governanceTreasurySafeForParent, vaultTreasurySaf
...
               onOpenCrewBootstrap={() => selectDashboardView('governance')}
+              onSelectNetwork={() => {
+                selectDashboardView('settings');
+                requestAnimationFrame(() => {
+                  document.getElementById('squad-settings-network')?.scrollIntoView({
+                    behavior: 'smooth',
+                    block: 'nearest',
+                  });
+                });
+              }}
```
logicminds: **[P1 blocking]** The Status checklist "Select network" CTA silently no-ops.

Previously this set `editingNetwork = true` and scrolled to the inline network row — one click landed in the `<select>`. The replacement switches to the Settings tab and, inside a single `requestAnimationFrame`, calls `document.getElementById('squad-settings-network')?.scrollIntoView(...)`. `DashboardSettingsTab` is lazy-loaded via `{#await loadDashboardSettingsTab() then ...}` (`createLazyComponent`, a real dynamic `import()`), so on the first click `#squad-settings-network` doesn't exist yet and the scroll is a silent no-op — even on the cached path the `.then` + render flush can lose the race with the rAF. Nothing puts `SquadNetworkSection` into edit mode anymore either, so the user lands on Settings and must still find the pencil button manually.

Repro: squad with no network selected, click the checklist CTA — the primary onboarding action appears to do nothing.

src/lib/squad/squad-identity-announce.ts
```diff
@@ -0,0 +1,85 @@
...
+    return { parent_id, icon_url: normalizeIconUrl(p.icon_url) };
+  } catch {
+    return null;
+  }
+}
+
+function normalizeIconUrl(value: unknown): string | null {
+  if (typeof value !== 'string') return null;
+  const trimmed = value.trim();
+  return trimmed.length > 0 ? trimmed : null;
```
logicminds: **[P2]** `normalizeIconUrl` accepts any non-empty string. `parseSquadIdentityUpdated` hands `applySquadIdentityUpdated` an arbitrary remote-peer value that gets persisted to SQLite and rendered directly as `` in `SquadAvatar.svelte:44` with no filtering — despite `docs/communities/SQUAD_CATALOG.md:16` specifying "Blossom HTTPS avatar". Any squad member can point every other member's client at an attacker-controlled URL (`http://` tracking pixel leaking IP/UA on every render, oversized asset, or a `data:` payload). The codebase already has this guard for remote avatars: `isHttpUrl` in `src/lib/utils/profile.ts:17-20`, used by `getProfileAvatarSrc`. Reject non-https values here (or at the apply boundary). Same gap applies to the `iconUrl` carried on `squad_invite` (`src/lib/parent/invite-members-flow.ts:78`).

src/lib/squad/squad-identity-announce.ts
```diff
@@ -0,0 +1,85 @@
...
+}
+
+function sameGroupId(a: string, b: string): boolean {
+  return a.trim().toLowerCase() === b.trim().toLowerCase();
+}
+
+export async function publishSquadIdentityUpdated(parent: Squad): Promise<boolean> {
+  const announcements = getAnnouncementsChannel(parent);
+  const gid = announcements.groupId?.trim() || parent.id.trim();
+  if (!gid) return false;
```
logicminds, line 61: **[P2]** Unsafe `getAnnouncementsChannel(parent).groupId?.trim()` — `getAnnouncementsChannel` returns `find(...) ?? sortedChannels[0]`, which is `undefined` for a parent with no channels despite the return type (the codebase acknowledges this elsewhere: `getAnnouncementsChannel(parent)?.groupId?.trim()` in `parent-navbar.ts:96`). Channel-less squads exist in the store (squad-pair drafts in `squad-pair-create.ts:138-142`, the optimistic row in `Navbar.svelte:373-374`). The same unsafe form is used in `squad-state-sync.ts:266`. Since this runs before every republish, a `TypeError` here aborts the *entire* sync response (EVM, infra, network, RPC, channels catalog) as an unhandled rejection, not just identity. Add the `?.` on the channel itself.

src/lib/squad/squad-identity-announce.ts
```diff
@@ -0,0 +1,85 @@
...
+  if (!sameGroupId(parsed.parent_id, groupId)) return;
+  const parent = get(squads).find(
+    (s) => sameGroupId(s.id, parsed.parent_id) || sameGroupId(s.id, groupId),
+  );
+  if (!parent) return;
+  const nextUrl = parsed.icon_url ?? undefined;
+  const current = parent.iconUrl?.trim() || undefined;
+  if (current === nextUrl) return;
+  void persistSquadPatch(parent.id, (squad) => ({ ...squad, iconUrl: nextUrl }));
+}
```
logicminds, line 86: **[P2]** `applySquadIdentityUpdated` fires `void persistSquadPatch(...)` unsequenced. `persistSquadPatch` updates the store synchronously then awaits `upsert_squad` (`squad-catalog.ts:151-159`), so two announces arriving within one DB round-trip start two concurrent upserts with no guaranteed completion order — the store can show the newer icon while SQLite retains the older one, reappearing on the next `hydrateSquadsFromDb`. Not hypothetical: the sync-response bug above means a joiner can receive N near-simultaneous announces. The payload also carries no `updated_at_ms` (the catalog row has one), so there's no way to reconcile — "last received wins" is enforced only by arrival order in the store, not in persistence. Await/queue the patch per parent, or add a monotonic timestamp and drop older values.

src/components/squad/PairWithSquadModal.svelte
```diff
@@ -82,14 +84,22 @@
...
+      <AvatarPicker
+        src={iconUrl || null}
+        editable
+        allowClear={!!iconUrl}
+        size={72}
+        chooseTitle={$t('squad.pfp.choose')}
+        editAriaLabel={$t('squad.pfp.change')}
+        cropTitle={$t('squad.pfp.cropTitle')}
+        onChange={(url) => (iconUrl = url)}
+        onClear={() => (iconUrl = '')}
```
logicminds, line 97: **[P1 dup, same root cause]** Same nested-`Modal`-Escape bug as `AvatarPicker.svelte` — see that file's top-level comment. This call site (Pair with Squad) discards the entered name/partner/visibility/tags on Escape-to-cancel-crop, and defeats `dismissible={!uploading}`.

src/components/layout/Navbar.svelte
```diff
@@ -712,14 +713,22 @@
...
+        chooseTitle={$t('squad.pfp.choose')}
+        editAriaLabel={$t('squad.pfp.change')}
+        cropTitle={$t('squad.pfp.cropTitle')}
+        onChange={(url) => (organizeSquadIconUrl = url)}
+        onClear={() => (organizeSquadIconUrl = '')}
+      >
+        {#snippet fallback()}
+          <SquadAvatar src={null} name={organizeSquadName} seed={organizeSquadName} fill />
+        {/snippet}
+      </AvatarPicker>
```
logicminds, line 731: **[P1 dup, same root cause]** Same nested-`Modal`-Escape bug — see `AvatarPicker.svelte`'s comment. This call site (Organize/Create Squad) is reset entirely by `openOrganizeSquadModal()` on next open, so the discarded state is a real loss for the user.

src/components/parent/dashboard/SquadIdentitySection.svelte
```diff
@@ -0,0 +1,83 @@
...
+    saving = true;
+    saveError = null;
+    try {
+      const patched = await persistSquadPatch(squad.id, (row) => ({
+        ...row,
+        iconUrl: url?.trim() || undefined,
+      }));
+      if (patched) {
+        await publishSquadIdentityUpdated(patched);
+      }
```
logicminds, line 32: **[P3]** `saveIcon` awaits `publishSquadIdentityUpdated` but discards its boolean return. That function never throws — it catches send errors, logs via `dmError`, and returns `false` (also `false` when no announcements group id resolves). So when the MLS announcement fails, the local catalog row is already patched and the new avatar renders locally with no error shown; other members silently keep the old photo with no retry affordance. Check the return value and surface a toast on `false`.

src/components/parent/dashboard/DashboardStatusTab.svelte
```diff
@@ -24,56 +17,59 @@
...
+      crewWearers,
+    }),
+  );
+
+  $effect(() => {
+    const pid = parentId;
+    const gid = announcementsGroupId;
+    if (!pid) return;
+    void needsSquadRosterKeyChoice(pid, gid).then((needed) => {
       rosterKeyNeeded = needed;
```
logicminds, line 74: **[P3]** `onMount` was replaced with an `$effect` that re-runs on `parentId`/`announcementsGroupId` changes, but the `needsSquadRosterKeyChoice` result is applied unconditionally (`rosterKeyNeeded = needed`) with no cancellation guard. Switching squads can let an older in-flight call overwrite a newer answer, masking/revealing roster keys for the wrong squad. Same pattern duplicated in `DashboardSettingsTab.svelte:57-64`. Elsewhere in this file the convention is a captured-id guard (`if (parentId !== loadParentId) return;` in `loadSquadMemberEvm`) — apply the same here.

src/components/settings/ProfileSection.svelte
```diff
@@ -175,29 +150,17 @@
...
+            src={isEditing ? (editAvatarUrl || avatarSrc) : avatarSrc}
+            editable={isEditing}
+            disabled={savingProfile}
+            size={128}
+            chooseTitle={$t('profile.chooseAvatarImage')}
+            editAriaLabel={$t('profile.changeAvatar')}
+            cropTitle={$t('profile.crop.title')}
+            fallbackText={(profile.display_name || profile.name || $t('profile.avatarFallback')).charAt(0)}
+            onChange={handleAvatarChange}
+          />
```
logicminds, line 164: **[P3]** Accessibility regression: the removed markup rendered ``. `AvatarPicker` hardcodes `alt=""` on its `` (`AvatarPicker.svelte:81`) and marks the letter fallback `aria-hidden`, with no prop to override the alt text — the user's own avatar is now announced as purely decorative. AvatarPicker needs an optional `alt`/`imageAlt` prop threaded through from here.

src/components/parent/governance/SmartContractSecuritySection.svelte
```diff
@@ -396,14 +396,24 @@
...
+    background: var(--bg-elevated);
+    border: 1px solid var(--border-subtle);
+    border-radius: 10px;
   }
 
   .smart-contract-security--compact {
-    margin-top: 8px;
-    padding-top: 8px;
+    margin: 0;
+  }
```
logicminds: **[P3]** `.smart-contract-security` lost `margin-top`/`padding-top`, so `.smart-contract-security--compact`'s sole remaining declaration `margin: 0` now overrides nothing — leftover from the pre-card layout, safe to delete.

src/components/parent/dashboard/SquadStickersSection.svelte
```diff
@@ -188,12 +188,12 @@
   }
 </script>
 
-<section class="stickers-tab" aria-label={`${$t('squad.stickers.tabAria')} — ${squad.name}`}>
+<section class="stickers-section" aria-label={`${$t('squad.stickers.sectionAria')} — ${squad.name}`}>
+  <h3 id="squad-stickers-heading" class="stickers-title">{$t('squad.stickers.sectionTitle')}</h3>
```
logicminds: **[P3]** `id="squad-stickers-heading"` has no consumer — the section names itself via `aria-label` rather than `aria-labelledby`, and there's no other reference to this id repo-wide. The sibling card in the same host (`SmartContractSecuritySection`) uses the paired pattern (`aria-labelledby` on the section, matching id on the heading) — either wire it the same way or drop the dangling id.

src/components/parent/dashboard/SquadBotHoldersSection.svelte
```diff
@@ -239,17 +239,17 @@
...
 <style>
   .squad-bot-holders-section {
-    margin-bottom: 16px;
+    padding: 14px 16px;
+    background: var(--bg-elevated);
     border: 1px solid var(--border-subtle);
-    border-radius: 8px;
-    padding: 16px;
+    border-radius: 10px;
   }
```
logicminds, line 246: **[P3]** This section gained `background: var(--bg-elevated)`, the same token `.bot-key-box` (line 283) already uses for its inner surface — the nested key box is now the same color as its card, distinguished only by a 1px border. Sibling cards (`SquadNetworkSection`, `SquadEndpointsPanel`) don't nest an `--bg-elevated` box inside an `--bg-elevated` card. Reads as an unnoticed side effect of the card conversion.

src/lib/mls/virtual-channel-bucket.test.ts
```diff
@@ -351,6 +351,15 @@ describe('announceCardAllowedForTimelineBucket', () => {
...
   });
 
+  it('routes squad_identity_updated to announcements', () => {
+    const raw = JSON.stringify({
+      type: 'squad_identity_updated',
+      payload: { parent_id: 'p', icon_url: 'https://cdn.example/a.jpg' },
+      pacto_virtual_bucket: 'announcements',
+    });
+    expect(deriveVirtualBucketFromMessageContent(raw)).toBe('announcements');
+  });
```
logicminds, line 360: **[P3]** This fixture sets `pacto_virtual_bucket: 'announcements'` explicitly, which `deriveVirtualBucketFromMessageContent` honors before reaching the `squad_identity_updated` branch — the function also defaults to `'announcements'` for unrecognized JSON, so this assertion passes even with the new branch deleted. The adjacent `squad_network_updated` test deliberately omits `pacto_virtual_bucket` for exactly this reason. Same issue in `src-tauri/src/virtual_channel_bucket.rs:238-243`. Drop the key from both fixtures so they actually exercise the new arm.

src/lib/app/mls-structured-refresh.test.ts
```diff
@@ -278,6 +278,15 @@ describe('onMlsStructuredMessage', () => {
...
       handlers,
     );
+    onMlsStructuredMessage(
+      JSON.stringify({
+        type: 'squad_identity_updated',
+        payload: { parent_id: 'g1', icon_url: 'https://cdn.example/a.jpg' },
+        pacto_virtual_bucket: 'announcements',
+      }),
+      'g1',
+      handlers,
```
logicminds: **[P3]** The added `squad_identity_updated` dispatch call has no accompanying expectation, and `../squad/squad-identity-announce` isn't mocked — `applySquadIdentityUpdated` runs against the real (empty) squads store and returns early at the parent lookup. The test still passes if the new dispatch block (`mls-structured-refresh.ts:104-106`) is deleted. Mock the module and assert it was called with `(raw, 'g1')`, matching how `respondToSquadStateSyncRequest` is asserted elsewhere in this file.

src/components/dm/InviteCard.svelte
```diff
@@ -57,7 +59,9 @@
 
 <div class="invite-card" class:collapsed role="article">
   <div class="invite-card-icon">
-    {#if inviterAvatarSrc}
+    {#if squadIconUrl}
+      <SquadAvatar src={squadIconUrl} name={squadName} seed={squadName} fill />
+    {:else if inviterAvatarSrc}
```
logicminds, line 65: **[P3]** Every other `SquadAvatar` call site seeds the fallback gradient with the squad's stable id (`ParentSidebar.svelte:100`, `PairWithSquadModal.svelte:114`, `SquadIdentitySection.svelte:53`). This seeds with `squadName` instead, so a squad whose icon URL fails to load shows a different gradient here than in the sidebar after joining. `DmMessageRouter` already has `presentation.payload.groupId` available to pass through.

src/lib/media/pick-local-image.ts
```diff
@@ -0,0 +1,18 @@
...
+}): Promise<string | null> {
+  const selected = await openFileDialog({
+    title: options.title,
+    filters: [{ name: options.filterName, extensions: [...IMAGE_EXTENSIONS] }],
+    multiple: false,
+  });
+  if (selected == null) return null;
+  if (typeof selected === 'string') return selected;
+  return selected[0] ?? null;
+}
```
logicminds, line 17: **[P3]** `open()` is called with `multiple: false`, so `OpenDialogReturn` resolves to `string | null` for this call. After the `null` and `typeof === 'string'` guards, `selected` narrows to `never` — `selected[0] ?? null` is dead code that only survives because indexing `never` isn't a type error. The code this replaces (`ProfileSection.handleChangeAvatar` on `origin/main`) assigned the result directly with no array handling.
```suggestion
  if (selected == null) return null;
  return selected;
}
```

src/components/ui/Tab.svelte
```diff
@@ -9,8 +9,11 @@
   export let hasUnreadDot = false;
 
   $: firstLetter = label.charAt(0).toUpperCase();
+  $: showImage = !!image && !imageBroken;
 
   let buttonEl: HTMLButtonElement;
+  let imageBroken = false;
+  $: image, (imageBroken = false);
   let showTooltip = false;
```
logicminds, line 17: **[P3]** `$: showImage = !!image && !imageBroken` is declared before `let imageBroken = false` and its reset (`$: image, (imageBroken = false)`). Svelte 5 legacy mode runs `$:` statements as pre-effects in source order, not topologically sorted like Svelte 4 — on an `image` change, `showImage` first recomputes against the stale `imageBroken` from the previous image, converging only because the reset re-dirties it and forces a second pre-effect pass in the same flush. Result is correct, but depends on effect re-entry rather than top-to-bottom reading order. Move `showImage` below the reset.

src/components/ui/AvatarCropModal.svelte
```diff
@@ -562,9 +535,6 @@
   }
 
   .crop-viewport-pending {
-    /* Not display:none: that would zero out clientWidth and break the cropDiameter binding
-       while the preview is still loading. visibility:hidden keeps layout (and measurement)
-       intact while hiding the not-yet-sized image from view. */
     visibility: hidden;
   }
```
logicminds, line 541: **[P3]** The rename from `settings/AvatarCropModal.svelte` stripped the comments documenting two non-obvious layout invariants: why `.crop-viewport-pending` uses `visibility: hidden` not `display: none` (the latter zeroes `clientWidth`, breaking the `bind:clientWidth={cropDiameter}` binding all pan/zoom math depends on), and why `.crop-viewport` uses `box-shadow` not `border` (a border changes the box model `cropDiameter` assumes). `e2e/avatar-crop.spec.ts:131` still says "see the CSS comment on `.crop-viewport` in AvatarCropModal.svelte" — that comment no longer exists and the file moved out of `settings/`. Restore a one-line note or point the e2e comment at a tracked doc.

docs/communities/SQUAD_CATALOG.md
```diff
@@ -13,7 +13,7 @@ Per-account squad and squad-pair rows live in **`squads`** (`pacto.db`), not `lo
 | `commons_tags` | Optional last-used tag defaults (exactly 3 when set). Tags for an active broadcast are on the Nostr event. |
 | `paired_squads` | JSON `[{id,name},{id,name}]` when squad-pair |
 | `channels` | JSON `[{name, groupId, order}]` |
-| `icon_url` | Optional |
+| `icon_url` | Optional Blossom HTTPS avatar. Set via the shared image picker at Organize Squad / pair create, or Dashboard → Status. Members sync through MLS `squad_identity_updated` on `#announcements` (last received wins) and optional `iconUrl` on `squad_invite`. Blank clears. |
```
logicminds: **[P2]** "Optional Blossom HTTPS avatar" only holds for locally-set icons (upload → `AvatarCropModal` → Blossom). Values arriving over the wire (`squad_identity_updated`, `squad_invite.iconUrl`) are accepted as any non-empty trimmed string with no scheme check (`normalizeIconUrl` in `squad-identity-announce.ts:46-50`, `parseSquadInviteMessage` in `nostr.ts:567-568`) and rendered directly as `` by `SquadAvatar`. Either qualify this row (HTTPS expected, not enforced on received values) or add the validation — see the inline comment on `squad-identity-announce.ts`.

src/lib/squad/squad-state-sync.ts
```diff
@@ -260,6 +261,10 @@ export async function respondToSquadStateSyncRequest(
   const wantRpc = !req.requested?.length || req.requested.includes('rpc');
   const wantChannels = !req.requested?.length || req.requested.includes('channels');
 
+  const parent =
+    get(squads).find((s) => s.id === parentId) ??
+    get(squads).find((s) => getAnnouncementsChannel(s).groupId === parentId);
```
logicminds: **[P2]** `formatSquadStateSyncRequest` adds `'identity'` to `requested` (line 105) but nothing reads it here — `wantEvm/wantInfra/wantNetwork/wantRpc/wantChannels` are computed above, then identity is published unconditionally below under `if (parent)` (line ~340). Consequences: (1) `'identity'` is dead protocol surface; (2) a peer requesting a narrower subset still gets an identity announce, including the destructive-clear case flagged on the identity republish comment; (3) `anyOk` can flip `true` purely from the ungated identity publish. Add `const wantIdentity = !req.requested?.length || req.requested.includes('identity')` and gate the block. Also `docs/communities/DESIGN.md:117` still documents `requested: ['evm','infra','network','channels']` — missing both `'rpc'` and `'identity'`.

src/lib/invites/accept-invite.ts
```diff
@@ -277,6 +277,7 @@ export interface AnnouncementsInvitePayload {
   groupId: string;
   name: string;
   memberSquads?: { id: string; name: string }[];
+  iconUrl?: string;
```
logicminds, line 280: **[P2]** New `iconUrl` field on `AnnouncementsInvitePayload` only reaches `finalizeSquadAfterAnnouncementsWelcome` when a Welcome already exists or arrives within `ACCEPT_WELCOME_FAST_PATH_MS`. In the common consent-first path (admitter runs MLS Add later), only `{messageId, groupId, squadName, inviteId, acceptedAt}` is persisted via `upsertPendingSquadAdmission`; `tryCompletePendingSquadAdmission` then finalizes with `{groupId, name: pending.squadName}` — no icon field — so the new squad is created and persisted without its photo until a post-join state-sync identity republish backfills it (best-effort, requires a responder online). Add optional `iconUrl` to the pending admission record and thread it through `tryCompletePendingSquadAdmission`.

src/components/parent/dashboard/SquadEndpointsPanel.svelte
```diff
@@ -74,7 +74,7 @@
   );
 
   $effect(() => {
-    const nonce = $squadStatusRpcFocusNonce;
+    const nonce = $squadSettingsRpcFocusNonce;
```
logicminds, line 77: **[P3]** The rename pass (`squadStatusRpcFocusNonce`→`squadSettingsRpcFocusNonce` here, `focusSquadStatusRpcEditor`→`focusSquadSettingsRpcEditor`, `squad-status-network`→`squad-settings-network` elsewhere) missed this panel's scroll target — the anchor id at line 83 below (`document.getElementById('squad-status-rpc')`) is still `squad-status-rpc` even though the panel now lives only under Settings. Cosmetic (self-referential within this file, still resolves), but inconsistent with the rest of the rename.

src/lib/i18n/locales/en/profile.json
```diff
@@ -31,21 +31,5 @@
 	"profile.couldNotPublishProfile": "Could not publish profile.",
 	"profile.failedUploadAvatar": "Failed to upload avatar",
```
logicminds: **[P3]** `profile.imagesFilter` (line 29, unchanged above this hunk) and `profile.failedUploadAvatar` (line 32) have no remaining `$t()` call site — the crop modal migrated to `media.avatarCrop.imagesFilter`, and `ProfileSection.handleChangeAvatar` (the only consumer of `failedUploadAvatar`) was deleted in this patch. `profile.uploading` (line 12, not in this diff) is orphaned the same way. Same three keys in `es/profile.json`. The parity test still passes (both locales keep them) but they're dead translation surface. `profile.crop.title`/`profile.crop.guidance` are still used and should stay.

src/components/parent/dashboard/SquadBroadcastSettingsSection.svelte
```diff
@@ -183,16 +183,8 @@
...
-          class:is-active={hasActive}
-          style={squad.iconUrl ? '' : `background-image: ${commonsTagGradient(squad.id || squad.name)}`}
-        >
-          {#if squad.iconUrl}
-            <img src={squad.iconUrl} alt="" loading="lazy" decoding="async" />
-          {:else}
-            <span aria-hidden="true">{squad.name.charAt(0).toUpperCase()}</span>
-          {/if}
+        <div class="commons-personal-avatar" class:is-active={hasActive}>
+          <SquadAvatar src={squad.iconUrl} name={squad.name} seed={squad.id || squad.name} fill />
```
logicminds, line 187: **[P3]** Swapping to `<SquadAvatar ... fill />` here leaves dead CSS below in the `<style>` block: `.commons-personal-avatar img` and `.commons-personal-avatar span` (around line 358+) no longer match anything since the ``/`<span>` moved into the child component. `svelte-check` will flag `css_unused_selector` on both — this file compiles warning-free on `origin/main`. The `background-size`/`background-position` declarations on `.commons-personal-avatar` are also dead now that no inline `background-image` is set.

src/components/layout/ParentSidebar.svelte
```diff
@@ -274,7 +278,7 @@
     font-size: 1rem;
     font-weight: 600;
     color: var(--text-primary);
-    margin: 0 0 4px 0;
+    margin: 0;
     overflow: hidden;
```
logicminds, line 282: **[P3]** `.parent-name`'s `margin: 0` (was previously `margin-bottom: 4px`) removes the only separation from `.parent-subheading` (`margin: 0` too, just below). Correct for centering the name against the new 32px `SquadAvatar`, but for squad-pair hubs (the only case `subheading` is set, see `ParentNavbar.svelte:111-116`) the partner-squad line now sits flush against the squad name. Add `margin-top: 4px` on `.parent-subheading` or a `gap` on the containing `.parent-heading`.

src/components/commons/CommonsBroadcastCard.svelte
```diff
@@ -133,8 +134,10 @@
   on:click={handleCardClick}
   on:keydown={handleCardKeydown}
 >
-  <div class="commons-tile-cover" style={coverImage ? '' : `background-image: ${commonsTagGradient(coverSeed)}`}>
-    {#if coverImage}
+  <div class="commons-tile-cover" style={isSquad || coverImage ? '' : `background-image: ${commonsTagGradient(coverSeed)}`}>
```
logicminds, line 137: **[P3]** `coverImage` (computed a few lines up from `broadcast.squadIconUrl` for squads) is only read here via `isSquad || coverImage`, which short-circuits to `''` whenever `isSquad` is true — so the squad branch of `coverImage` is never actually evaluated for its value. Same dead branch in `CommonsBroadcastDetailModal.svelte:51`. Reducing `coverImage` to `getProfileAvatarSrc(userProfile)` (drop the squad ternary) keeps behavior identical and removes the misleading impression that `coverImage` still drives squad tiles — `SquadAvatar` on the next line does that job now.

logicminds (changes requested):
## Code review summary — squad PFP picker

8 parallel reviewers covered the diff by locality; 29 inline comments posted on the specific lines involved. Top-level summary below; full detail is inline.

### Blocking (P1)

1. **Nested `Modal` + Escape key discards form state** — `AvatarPicker` mounts `AvatarCropModal` (a full `Modal`) inline. Used inside another `Modal` at `Navbar.svelte:717-731` (Organize/Create Squad) and `PairWithSquadModal.svelte:88-97`, one Escape press closes both the crop dialog and the enclosing form, discarding all entered state, and defeats `dismissible={!uploading}` mid-upload. Fix once in `Modal.svelte`'s `handleContentKeydown` (`stopPropagation()` on a handled Escape).
2. **`squad_identity_updated` sync-response can wipe the squad photo squad-wide** — `respondToSquadStateSyncRequest` always republishes `icon_url: parent.iconUrl ?? null`; a member with no local icon broadcasts a clear that every receiver applies. Routine trigger: any new joiner's sync request.
3. **Dashboard "Select network" checklist CTA silently no-ops** — the new scroll target lands in a lazy-loaded tab (`createLazyComponent`) that hasn't rendered yet on first click; nothing enters edit mode anymore either.

### Should fix (P2)

- No scheme validation on peer-supplied `icon_url` (`squad_identity_updated` and `squad_invite.iconUrl`) — rendered directly as `` with no `isHttpUrl`-style guard, despite docs claiming HTTPS-only.
- `getAnnouncementsChannel(...).groupId` used unguarded in two new call sites — can throw and abort the *entire* sync response (EVM/infra/network/RPC/channels), not just identity.
- `AvatarPicker`'s "Remove photo" control disappears exactly when it's needed (image load failure nulls `displaySrc`, which also gates `showClear`) — reachable via `SquadIdentitySection`.
- `'identity'` requested-capability token added to the sync-request protocol but never read/gated on the responder side.
- No persistence-side ordering for "last received wins" — concurrent unsequenced `persistSquadPatch` calls can leave stale data in SQLite with no timestamp to reconcile.
- Deferred (consent-first) squad admission path drops `iconUrl` entirely — joiners see no photo until a best-effort resync backfills it.

### Nits (P3)

Rename miss (stale `squad-status-rpc` anchor id after the Settings-tab move), two unguarded async races in roster-key effects, a swallowed `publishSquadIdentityUpdated` failure with no user-facing toast, several orphaned i18n keys (`profile.imagesFilter`/`profile.failedUploadAvatar`/`profile.uploading`), a lost `alt` text on the profile avatar, dead CSS left behind by the `SquadAvatar` swap in three files, a dangling unused heading id, vacuous new tests for `squad_identity_updated` bucket routing (both TS and Rust) and dispatch, an inherited squad-id-vs-announcements-group-id matching gap, and a few cosmetic gradient-seed/spacing inconsistencies. All itemized inline.

### What's solid

Payload shape (publish↔parse↔apply) is consistent; TS/Rust bucket-routing lists both correctly learned `squad_identity_updated`; structured-notice i18n key + en/es parity exact (356/356 keys verified programmatically); own-echo is idempotent; Commons publish-time icon snapshot semantics match documented behavior; invite `iconUrl` threading is type-correct end-to-end with clean backward compat for pre-iconUrl invites; the Status→Settings dashboard section move is a faithful 1:1 relocation with no dropped markup; the `DashboardStickersTab`→`SquadStickersSection` rename is complete repo-wide; new files correctly use Svelte 5 runes per repo convention.

**Verdict: request changes** — the two functional P1s (photo-wipe race, dead-end onboarding CTA) and the UX-destroying P1 (modal Escape) should land before merge; the P2 security/robustness items are worth a fast follow-up if not bundled in.

daopunk:
`406abcfc` (`fix: address pr review`) lands the follow-up for @logicminds's requested changes.

**P1**
- Escape on the crop dialog no longer closes Organize / Pair Squad (including mid-upload).
- Sync responders no longer announce `icon_url: null`; identity republish only runs when the local photo exists and `requested` includes `identity`.
- Status “Select network” deep-links into Settings, waits for the lazy tab, and opens the network editor.

**P2**
- Peer `icon_url` / invite `iconUrl` must be `https://`.
- `getAnnouncementsChannel` is optional-chained so a channel-less squad cannot abort the whole sync response.
- “Remove photo” stays visible when the image fails to load.
- `persistSquadPatch` is serialized per squad.
- Consent-first pending admission keeps `iconUrl` through finalize.

**P3** nits in the same commit (toasts, roster-key race, profile `alt`, CSS/i18n leftovers, tests that actually hit the new identity arm).

logicminds (approved):

PR #281, opened 2026-08-18
fix: dm relay bug and error messaging in ui

### LINKED ISSUE #158: bug: error message in dms
## Summary
DMs show “Could not deliver to relays” error, but the error disappears after leaving and returning to the chat.

## Environment
- Device: MacBook
- App: Downloaded Pacto app
- Wallet/network: None
- Pacto URL/version: Unknown

## Steps to reproduce
1. Open a DM chat in the downloaded Pacto app.
2. Send or view messages in a DM with plor.
3. Observe the error: “Could not deliver to relays. Message may appear as pending or failed.”
4. Repeat or observe the same issue in a DM with trigs.
5. Leave the DM chat.
6. Return to the same DM chat.

## Expected behavior
If a message cannot be delivered to relays, the app should show an accurate persistent delivery state until the message is delivered, retried or confirmed failed.

If the message was actually delivered, the error should clear automatically without needing to leave and re-enter the DM chat.

## Actual behavior
The app displayed this error in multiple DM chats:

> Could not deliver to relays. Message may appear as pending or failed.

The error disappeared after leaving the DM chat and returning.

## Impact
This creates uncertainty about whether DMs were delivered. For coordination use cases, users need to know whether a message is pending, failed or successfully sent.

## Evidence
Observed in DMs with plor and trigs. No screenshot/version available.

daopunk:
## Summary

DMs could not tell you whether an outbound message actually reached a relay. After a send failed, the only signal was a thread banner — *Could not deliver to relays* — that vanished as soon as you left the chat and came back. The message itself looked sent.

That banner was session-only (`dmSendError` is cleared on `activeDmId` change). Real delivery state already lives on the message (`pending` / `failed` in SQLite) but the bubble never rendered it.

After this change, your own messages show **Sending…** while publish is in flight and **Not delivered** if relays never ACK. That status survives leave/reopen. The thread banner is reserved for invoke exceptions (invalid npub, client not initialized, upload errors), not for `ok === false`.

Closes #158.

## Changes

- Render per-message delivery copy on DM and channel bubbles (including compact stacked rows).
- Stop mapping backend `false` to the generic deliver-to-relays banner.
- Type `get_message_views` rows with `pending` / `failed` so reopen loads keep the flags.

## Test plan

- [ ] `pnpm exec vitest run src/lib/dm/resolve-dm-message-presentation.test.ts src/lib/api/nostr.test.ts`
- [ ] Send a DM, confirm **Sending…** appears on the bubble (no thread banner).
- [ ] Force a relay failure, confirm **Not delivered** in danger color, leave the chat, and reopen — the same status remains.
- [ ] Invalid npub / send exception still shows the thread banner.

---


logicminds (commented):
## Code review — approve with comments

Reviewed via two parallel passes (UI layer: ChatView/Message/MessengerChatView/+page.svelte/i18n; logic layer: nostr.ts, resolve-dm-message-presentation.ts + tests). **No blocking defects.**

### Verified non-issues
- `wallet.couldNotDeliverRequest` / `wallet.couldNotSendPaymentRequest` (used in `+page.svelte`'s payment-request path) exist at `en/wallet.json:83-84` and `es/wallet.json:80-81` — not in this diff because they're pre-existing keys outside its scope.
- New `errors.dm.*` / `messaging.message.*` keys present in both `en`/`es` with parity; removed `messaging.dm.newChat.deliverError` has zero leftover references.
- `get`/`t` imports in scope everywhere the new `tFn(...)`/`get(t)(...)` calls land.
- `outboundDeliveryLabel`'s `tFn: MessageFormatter = get(t)` default is re-evaluated per call (correct semantics), not a stale module-load snapshot.
- `--danger`/`--text-muted` are established theme tokens, defined in all five themes.
- Header vs. compact delivery badges are mutually exclusive (`{#if !compact}` / `{#if compact && deliveryLabel}`).
- **Banner-removal safety**: dropping the `if (!ok) $dmSendError = ...` path is safe end-to-end — backend flips `failed` and emits `message_update` on both the NIP-17 (`message.rs:422-455`) and MLS (`mls.rs:2862-2994`) paths, `tauri-subscriptions.ts` preserves `failed` through `normalizeDmPayload`, and the store replaces the row live, so the badge updates without a reload.
- `MessageViewDto` shape matches `DmMessage` exactly; no dead exports.

### Non-blocking findings (inline below, all P3)
1. `outboundDeliveryLabel` isn't gated on `msg.mine` despite being documented as outbound-only — no live defect today (received rows always persist `pending`/`failed` as 0), but a landmine for any future backend path that sets those columns on an inbound row.
2. Newly-visible `pending` badge can go permanently stuck if a crash/force-quit leaves a row `pending = 1` in SQLite — replays as "Sending…" forever with no retry UI yet (tap-to-retry is a deferred follow-up, but worth the retry ticket accounting for this).
3. `role="status"` on a `<span>` created together with its text is inconsistently announced by screen readers, and `loadOlder` pagination can inject a burst of these live regions at once.
4. `.message-delivery.pending` CSS rule duplicates `.message-delivery`'s color — dead rule.
5. Missing trailing newline in the new test file — not lint-enforced here (no `eol-last`/Prettier), pure style nit.

None of these block merge; items 1 and 2 are worth a follow-up ticket or a one-line fix at your discretion.

src/lib/dm/resolve-dm-message-presentation.ts
```diff
@@ -162,3 +163,13 @@ export function buildPlainMessageProps(
   }
   return base;
 }
+
+/** Outbound relay delivery: failed wins over pending; otherwise no status copy. */
+export function outboundDeliveryLabel(
```
logicminds, line 168: **Non-blocking (P3):** `outboundDeliveryLabel` is documented "Outbound relay delivery" but nothing gates it on `msg.mine` — `buildPlainMessageProps` and `ChatView.svelte`'s message-mapping both forward `pending`/`failed` before the `mine` branch, and `Message.svelte:48` calls this for every rendered row. No live defect today (received rows always persist `pending`/`failed` as 0 on the backend), but it's a landmine if any future backend path sets those columns on an inbound row — the peer's message would render "Not delivered". Cheapest guard: gate on a `mine` prop passed through from the two prop builders.

src/components/channel/ChatView.svelte
```diff
@@ -281,6 +281,7 @@
       attachments: msg.attachments,
       previewMetadata: msg.preview_metadata,
       pending: msg.pending,
+      failed: msg.failed,
```
logicminds, line 284: **Non-blocking (P3):** wiring `pending` to a visible badge makes any row left with `pending = 1` in SQLite render "Sending…" forever. That state is reachable — backend writes the outbound message with `pending: true` before the send attempt and only clears it on the success/failure callback, so a crash/force-quit mid-send leaves a stuck row that replays as "Sending…" on every subsequent thread open. Tap-to-retry is deferred to a follow-up, but that follow-up (or a stale-pending-on-load fix) should account for this new regression risk.

src/components/dm/Message.svelte
```diff
@@ -336,6 +339,14 @@
       <div class="message-header">
         <span class="author-name">{authorName}</span>
         <span class="timestamp"><time datetime={timestamp}>{formatMessageTimestamp(timestamp)}</time></span>
+        {#if deliveryLabel}
+          <span
+            class="message-delivery"
+            class:failed
+            class:pending={pending && !failed}
+            role="status"
```
logicminds, line 347: **Non-blocking (P3):** `role="status"` on a `<span>` created together with its text is an implicit `aria-live="polite"` region inserted with its content already present — assistive tech announces this inconsistently across screen readers. Also, `loadOlder` can prepend a page of history at once, so a batch of previously-failed messages could inject a burst of "Not delivered" live regions simultaneously on pagination, not just on a fresh send. A plain `<span>` (status is already adjacent to the message it describes) avoids both issues.

src/components/dm/Message.svelte
```diff
@@ -681,6 +700,25 @@
     font-weight: 400;
   }
 
+  .message-delivery {
+    font-size: 0.75rem;
+    font-weight: 400;
+    color: var(--text-muted);
+  }
+
+  .message-delivery.pending {
```
logicminds, line 709: **Non-blocking (P3), nit:** `.message-delivery` already sets `color: var(--text-muted)`; `.message-delivery.pending` sets the identical value, so this rule has no effect on any render. Reads as an intentional pending-specific treatment that was never differentiated — delete it or give pending a distinct token.

src/lib/dm/resolve-dm-message-presentation.test.ts
```diff
@@ -392,4 +393,32 @@ describe('buildPlainMessageProps', () => {
...
+
+  it('returns sending copy when pending and not failed', () => {
+    expect(outboundDeliveryLabel({ pending: true }, tFn)).toBe('messaging.message.sending');
+  });
+
+  it('returns null when neither pending nor failed', () => {
+    expect(outboundDeliveryLabel({}, tFn)).toBeNull();
+    expect(outboundDeliveryLabel({ pending: false, failed: false }, tFn)).toBeNull();
+  });
 });
```
logicminds, line 424: **Non-blocking, style nit:** file is missing a trailing newline (confirmed via `od -c`). Not lint-enforced in this repo (no Prettier config, `eslint.config.js` doesn't include `eol-last`), so purely cosmetic.

logicminds (approved):

Issue #312, opened 2026-08-20
feat: publish and fetch MLS key packages on kind 30443, dual-published with 443


logicminds:
## Impact

Publishing and fetching MLS key packages only on the legacy kind **443** costs us two things:

1. **We cannot invite anyone outside Pacto.** Other NIP-EE clients publish on kind **30443**. On `wss://nostr.computingcache.com` there are 10 live kind-30443 key packages from other clients (`Amethyst`, and one `MDK/0.8.0`), newest `2026-08-12`. Every one of them is invisible to our fetch filters, so a squad invite to those users can never resolve a key package.
2. **We are living on borrowed time.** MDK 0.8 accepts 443 only as `MLS_KEY_PACKAGE_KIND_LEGACY`, with its own source comment scoping that to "through May 31, 2026" — a date already behind us (`mdk-core-0.8.0/src/key_packages.rs`). Whenever upstream drops legacy acceptance, every Pacto-published key package becomes unparseable and squad invites stop working fleet-wide.

## Current state

- Spec kind is `30443`; `443` is legacy: `mdk-core-0.8.0/src/constant.rs:13-19`.
- We publish 443 with MDK's `tags_443` set: `src-tauri/src/lib.rs:9136`.
- We fetch 443 in four places: `refresh_keypackages_for_contact` (`src-tauri/src/lib.rs:10243`), cached-key-package verification (`src-tauri/src/lib.rs:9093`), `MlsService::create_group` fallback-by-author (`src-tauri/src/mls.rs:639`), and the invite path (`src-tauri/src/mls.rs:1276`).
- **No Pacto release has ever referenced 30443** — `grep -c 30443` is `0` at every tag from v0.1.0 through v0.6.1, and all of them publish/fetch `Kind::MlsKeyPackage` (443). So this is a fleet-wide flag day, not an old-build-versus-new-build compatibility question: the day we publish only 30443, every installed Pacto stops finding key packages.
- MDK already hands us both tag sets in one call: `KeyPackageData { tags_30443, tags_443 }` (`mdk-core-0.8.0/src/key_packages.rs:31-36`, built at `:213-246`). The 30443 variant carries the `d` tag, so it is addressable/replaceable rather than append-only.

## Proposed change — transition, then sunset

1. Publish **both** kinds from the same `KeyPackageData`: 30443 with `tags_30443`, 443 with `tags_443`.
[...]

PR #313, opened 2026-08-20
fix: create squads when a member's key package is stale, and say so when creation fails

### LINKED ISSUE #311: fix: republish the device key package when the cached one no longer parses
## Impact

A member whose published MLS key package is in the pre-MIP-00/02 format cannot be added to any squad, and the app never repairs it for them — **not even after they upgrade Pacto**. The stale key package on the relay is a sticky artifact: our own unlock path deliberately refuses to replace it.

## What happens

`regenerate_device_keypackage(cache = true)` runs on unlock (`src-tauri/src/lib.rs:7901`) and after PIN setup (`src-tauri/src/lib.rs:7852`). Its cache branch (`src-tauri/src/lib.rs:9061-9118`):

1. reads the local index entry for (my npub, my device_id),
2. fetches **that event id** from the trusted relays,
3. if the event still exists → returns `cached: true` and publishes nothing.

It checks *existence*, never *validity*. MDK 0.8 requires an explicit `["encoding","base64"]` tag on kind:443 key packages (`mdk-core-0.8.0/src/key_packages.rs:330`, `src/util.rs:60-79`, no default, no hex fallback). Key packages published by Pacto **v0.5.2 and earlier** (MDK git rev `f46875e`) are hex with no `encoding` tag, so they can never be parsed by a current build — yet they keep short-circuiting the republish for as long as a relay serves them.

Forced republish exists on only three paths, all `cache = false`: an MLS store reset (`keypackage_refresh_required`), group creation, and welcome acceptance. A user who upgraded and has not created a group or accepted an invite since is still advertising their old-format key package from a fully current build.

## Evidence

- Verified live against the real events: a member with a legacy key package makes MDK's `create_group` fail with `Missing required encoding tag`; the same call with a current key package succeeds.
- Real contact's key package: kind 443, `2026-07-28`, hex content (688 chars), no `encoding` tag, no `client` tag, extensions `0x0003,0x000a,0x0002,0xf2ee`.
- Current-format comparison from another contact: `2026-08-18`, base64, `encoding: base64`, `client: MDK/0.8.0`, extensions `0x000a,0xf2ee`.
- Format break landed in `09a568a` (nostr/MDK upgrade), first shipped in **v0.5.3** (2026-08-06). v0.5.0–v0.5.2 are on the old side.
- Squad creation no longer *aborts* on such a member (it now skips them and reports why), but they remain uninvitable until they republish.

## Proposed change

In the `cache = true` branch, run the fetched cached event through `engine.parse_key_package(&event)` — the same validation `MlsService::create_group` now performs — and treat a parse failure as a cache miss so a fresh key package is published and indexed. Self-healing on the next unlock: no migration table, no new wire format, no user action, no dependency on the peer noticing anything.

## Acceptance

- An account whose relay key package lacks the `encoding` tag republishes on next unlock, and the local index points at the new event id.
- An account with a valid current key package still short-circuits — no republish churn on every unlock.
- A relay-unreachable unlock neither wipes the index entry nor spams publishes.
- Backend test: cached event exists but fails to parse → republish; cached event parses → no publish.

Related: MLS key package kind migration (443 → 30443).

logicminds:
## Summary

Creating a squad with a member whose published MLS key package predates the MDK 0.8 format did nothing at all: the squad flashed into the rail, vanished, and the only trace was a `console.error`. This PR fixes both halves of that — the creation no longer dies on one stale key package, and a creation that *does* fail now says so.

**Before:** click Create → the new squad appears for an instant, disappears, no message, no retry, nothing in the UI. The squad was never created and the user has no idea why.

**After:** members whose key package cannot be parsed are left out of the new squad and named in the toast, so the squad is created with everyone else. If creation genuinely fails, the placeholder stays in the rail with the error, a working Retry, and a Discard.

Root cause of the failure was MDK 0.8 requiring an explicit `["encoding","base64"]` tag on kind:443 key packages and parsing every member's key package in one `collect::<Result<_,_>>()?` — so a single member still advertising a pre-MIP-00/02 (hex, untagged) key package failed the whole group with `Missing required encoding tag`. Root cause of the silence was the frontend catch path deleting the optimistic placeholder from `squads` and then writing the error under that now-dead id, making the existing inline error and Retry button unreachable dead code.

Also fixes the reason such a key package never healed itself (#311): the unlock-path republish verified that the indexed key package still *existed* on a relay, never that it still *parsed*, so an account that published before upgrading kept advertising an unusable event forever and could not be invited anywhere.

## Changes

- `MlsService::create_group` validates each candidate key package with `parse_key_package` before handing them to the engine, creates the group with the survivors, and returns `GroupCreateOutcome { group_id, skipped }`. It fails only when nobody is left.
- `create_group_chat` returns `{ groupId, skippedMembers }`, merging preflight skips (no key package published) with engine-time skips (unfetchable or unparseable).
- `regenerate_device_keypackage`'s cache branch parses the fetched cached event and treats a parse failure as a cache miss, so the next unlock republishes in the current format.
- A failed squad create keeps its placeholder selected with the error, the existing Retry, and a new Discard; an error toast carries the same retry action. Skipped members are named in the ready toast and are no longer sent invite DMs they could not accept.

## Test plan

- `cargo test --lib` — 813 passing, including new coverage that a key package stripped of its `encoding` tag is rejected and the all-skipped error names each member.
- `pnpm test` — 2142 passing, including that a rollback keeps the placeholder and that skipped members are excluded from invite DMs.
- `pnpm check`, `pnpm lint` — clean.
- Live, two sandbox personas on a local relay:
  - squad create with `[modern member, legacy member]` → `{groupId: 4b316858…, skippedMembers: [{npub1el6w…, "Missing required encoding tag"}]}`; the same call previously errored outright.
  - legacy member alone → still errors, now naming the member and reason.
  - relay stopped mid-create → placeholder retained with error + Retry + Discard + toast; relay restored + Retry → squad finalized (`groupId 6fa214d4…`) and invite DM sent; Discard → placeholder gone from the rail and from persisted nav order.
  - planted a stale key package ref for the sandbox's own device → on next unlock: `Cached event 0cb5d6e7… no longer parses (Missing required encoding tag); republishing`, index repointed at a fresh event carrying `encoding: base64` / `client: MDK/0.8.0`; a second call returns `cached: true` with no republish churn.

## Follow-up

#312 tracks moving key packages to spec kind 30443 (dual-published with 443). That is what lets us invite non-Pacto NIP-EE users at all, and this PR's republish gate is the mechanism by which the fleet would acquire those key packages.

Closes #311

---


logicminds:
Follow-up filed as #321 — a failed MLS welcome delivery leaves an orphan group in the engine store, and every Retry mints another.

The retry latch in this PR stops accidental duplicates from double-clicking, but not this one: `create_group` returns `Err` on welcome-publish failure *before* `write_groups` runs, so the group exists in the local MDK store with no metadata, no chat, and no UI row. Partial delivery is the sharp edge — the members whose welcomes did publish hold a valid invite to a group the creator abandoned.

Deliberately out of scope here: it needs a decision about where the create boundary sits (persist-then-deliver vs. retry-adopts-existing), not a patch.

Issue #337, opened 2026-08-21
chore: sunset legacy kind:443 MLS KeyPackage publish/read after dual-publish transition


logicminds:
## Summary

#312 made Pacto dual-publish MLS KeyPackages on both kind **30443** (spec) and legacy kind **443**, and widened every fetch filter to accept either. That is a transition state, not the end state: per MIP-00 (and MDK's own `MLS_KEY_PACKAGE_KIND_LEGACY` doc comment, "through May 31, 2026"), kind 443 is meant to go away. This issue tracks turning off the legacy side once it's safe to.

## Two independent cutovers

1. **Stop *publishing* kind 443.** `regenerate_device_keypackage` (`src-tauri/src/lib.rs`) currently signs and sends two events per (re)publish. Once the fleet has had a release that dual-publishes for long enough that essentially every reachable device has re-published at least once, drop the second `send_event_to` call for the legacy event — publish 30443 only. The `keypackage_ref_secondary` column and the "both refs must independently verify" cache-gate logic in the same function should also collapse back to a single ref at this point, since there's no longer a second kind to keep alive.
2. **Stop *reading* kind 443** (separately, likely later). `mls::mls_key_package_kinds()` is the single choke point every fetch filter in `lib.rs`/`mls.rs` goes through (`refresh_keypackages_for_contact`, `MlsService::create_group`'s fallback-by-author, `MlsService::fetch_latest_keypackage`, and the cache-verification filter). Narrowing that helper to `[MLS_KEY_PACKAGE_KIND_30443]` and deleting `MLS_KEY_PACKAGE_KIND_LEGACY` retires reading in one place — but this must wait until no reachable device (Pacto or otherwise) still depends on Pacto reading its legacy-only KeyPackage.

These are separate go/no-go decisions: other NIP-EE clients (or an old Pacto install someone never reopened) may keep publishing 443-only for a while after Pacto itself stops. Reading legacy should outlive publishing legacy, mirroring MDK's own posture (`MLS_KEY_PACKAGE_KIND_LEGACY` stays *acceptable* long after publishers move on).

## Gate

- [ ] A telemetry-free way to gauge fleet coverage (e.g. sampling live relay data the way #312's original report did: "10 live kind-30443 key packages... newest 2026-08-12") shows kind 30443 adoption is effectively universal among reachable Pacto installs and the NIP-EE clients Pacto interops with.
[...]

PR #338, opened 2026-08-21
feat: dual-publish MLS key packages on kind 30443, accept legacy 443

### LINKED ISSUE #312: feat: publish and fetch MLS key packages on kind 30443, dual-published with 443

logicminds:
## Summary

Pacto only published and fetched MLS KeyPackages on the legacy Nostr kind 443. Every other NIP-EE client (Amethyst, MDK-based clients) publishes on the addressable spec kind 30443 instead, so a squad invite to any of them could never resolve a KeyPackage — and MDK's own upstream acceptance window for legacy 443 has already lapsed. This dual-publishes both kinds from the same KeyPackage content and widens every fetch path to accept either, so Pacto can invite and be invited by any NIP-EE client during the transition, while still resolving KeyPackages from installs that haven't upgraded.

## Changes

- `regenerate_device_keypackage` now signs and publishes both a kind:30443 (addressable, spec) and kind:443 (legacy) event from the same KeyPackage content, sharing one timestamp so the "prefer 30443" kind tie-break actually applies.
- The 30443 event reuses the device's stored `d` tag across rotations, so NIP-33 replacement actually retires the previous event on relays instead of leaving every past KeyPackage permanently live and addressable.
- Every fetch path — `create_group`'s member resolution, the invite path, `refresh_keypackages_for_contact`, and self-device cache verification — now matches both kinds and prefers the newest event the engine can actually parse, falling back to the newest candidate (parseable or not) so a genuine rejection reason still surfaces instead of a generic "not found."
- Self-device cache verification now tracks both event ids per device (`keypackage_ref_secondary`) and requires both to independently verify before trusting the cache, so one kind being pruned from a relay can't fool the check on the strength of the other.
- A publish every trusted relay rejects is now a hard error instead of being silently cached as if it succeeded.
- New migration adds `keypackage_ref_secondary` and `keypackage_d_tag` to `mls_keypackages`.

## Test plan

- `cargo test --lib` — 827 passed, 0 failed, including new golden-vector, fetch-filter, and sort-order tests.
- `cargo clippy --lib --no-deps` — no new warnings.
- Two migration test fixtures (baselined pre-existing account simulations) updated to include `mls_keypackages` so the new `ALTER TABLE` runs against a real table shape instead of failing with "no such table."

## Related

Fixes #312.
Related: #337 (follow-up — sunset legacy kind:443 publish/read once fleet adoption of 30443 is universal).

---

### DIFF
--- a/src-tauri/src/db.rs
+++ b/src-tauri/src/db.rs
@@ -4922,6 +4922,15 @@ mod legacy_mls_store_harvest_tests {
                 metadata TEXT NOT NULL DEFAULT '{}',
                 muted INTEGER NOT NULL DEFAULT 0
             );
+            CREATE TABLE mls_keypackages (
+                id INTEGER PRIMARY KEY AUTOINCREMENT,
+                owner_pubkey TEXT NOT NULL,
+                device_id TEXT NOT NULL,
+                keypackage_ref TEXT NOT NULL,
+                fetched_at INTEGER NOT NULL,
+                expires_at INTEGER NOT NULL
+            );
+            CREATE INDEX idx_keypackages_owner ON mls_keypackages(owner_pubkey);
             CREATE TABLE events (
                 id TEXT PRIMARY KEY,
                 kind INTEGER NOT NULL,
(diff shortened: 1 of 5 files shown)


daopunk (approved):
lgtm!

PR #339, opened 2026-08-21
fix: stop mls welcome-delivery failures from orphaning groups

### LINKED ISSUE #321: fix: a failed MLS welcome delivery leaves an orphan group, and every retry makes another
## Impact

A squad create whose welcome delivery fails leaves a **real MLS group in the local MDK store that the app never persists and never shows**, and every Retry mints another one. If the failure was partial — some welcomes published, others not — the members who did receive one hold a valid welcome for a group the creator abandoned. Accept it and you are alone in a group the creator's UI does not list and nobody else will ever join.

This is now easy to hit. Before #313 a failed create vanished silently and users gave up; #313 gave that failure a Retry button, and #313's follow-up made Retry reliable. Each press repeats the whole flow from scratch.

## Where

`MlsService::create_group` (`src-tauri/src/mls.rs`):

1. `engine.create_group(...)` succeeds — the group and its epoch-1 commit are now in the local MDK store (~line 788).
2. Welcomes are gift-wrapped to each invited recipient (~line 872).
3. Any failure returns early:
   ```rust
   if invited_count > 0 && welcome_failures > 0 {
       return Err(MlsError::NetworkError(format!(
           "Failed to deliver MLS welcome to {} of {} members", welcome_failures, invited_count
       )));
   }
   ```
   Same for the `welcome_rumors.is_empty()` branch below it.
4. `write_groups` / `mls_groups` metadata, the `STATE` chat row, and the `mls_group_initial_sync` emit all happen **after** that early return (~lines 924-985) and never run.

So the engine has a group with no metadata, no chat, and no UI presence. Nothing reconciles that direction: the only reference to cleanup is a stale doc comment in `lib.rs` (`/// Remove orphaned MLS groups from metadata that are not in engine state`) with no function under it, and it describes the opposite direction anyway.

## Reproduce

1. Create a squad with at least one member whose key package resolves.
2. Make welcome publication fail while the earlier fetches succeed — a relay that accepts REQs but rejects EVENTs, or a relay killed in the window between key-package fetch and gift-wrap.
3. Observe: create fails with `Failed to deliver MLS welcome to 1 of 1 members`, the placeholder stays with Retry, and `mls_groups` has no row.
4. Press Retry. A second group is created in the engine.

Inspect the store directly to see the orphans accumulate:
```sql
-- <app_data>/<npub>/mls/pacto-mls.db
select hex(mls_group_id) from groups;
```
and compare against the app's persisted set (`mls_groups`, per-account `pacto.db`).

## Why it needs a decision, not a patch

The current code treats "group exists locally" and "everyone can join it" as one atomic step, and they are not. Two coherent fixes, and they have different consequences:

**A. Persist before delivery, retry only delivery.** Write `mls_groups` metadata and the chat row immediately after `engine.create_group` succeeds, then attempt welcomes. A delivery failure surfaces as a squad that exists with members still pending, and Retry re-sends welcomes for the *existing* group instead of building a new one.
- No orphans, and the partial-success case becomes correct rather than merely non-destructive: the members who did get a welcome are joining the group the creator kept.
- Costs a new state the UI must express — "squad created, N members not yet invited" — and a re-send path keyed to an existing group id.

**B. Keep the current boundary; make retry adopt.** Leave the early return, but have retry look for an existing engine group matching this create's member set and resume from step 2 rather than calling `create_group` again.
- Smaller UI surface.
- Matching "the same squad" by member set is a heuristic, and it does nothing for the member who already accepted a welcome for the abandoned group.

**A** is the honest fix; **B** is the cheap one. Either way, decide before touching retry idempotency.

## Also needed regardless of which is chosen

- **Reap orphaned engine groups.** Groups in the MDK store with no `mls_groups` row are dead weight today and will exist in the wild from every failure that already happened. Reconcile on startup or on store-reset, and delete the stale doc comment in `lib.rs` that claims this already exists.
- **Test the boundary.** No test drives a create whose welcome publication fails. Add one that asserts exactly what survives a delivery failure — under fix A, that the group is persisted and the undelivered members are recorded; under fix B, that a retry adopts rather than creates.

## Not this issue

- Duplicate groups from *impatient clicking* — fixed by the per-parent retry latch in #313.
- Relay fetch failures during key-package resolution — fixed by the transient-skip abort in #313, which returns before the engine runs and so leaves nothing behind.

This issue is only the case where the engine already created the group and delivery is what failed.

logicminds:
## Summary

Before this fix, a squad/channel create whose welcome delivery failed silently orphaned a real MLS group: `MlsService::create_group` committed the group into the local engine store and then discarded it entirely on any delivery failure, leaving a group with no metadata, no chat, and no UI trace - retryable only by creating yet another orphan. Members who did receive a welcome ended up alone in a group the creator's UI never listed.

`create_group` now persists group metadata *before* attempting delivery, so a failed welcome becomes a resendable "pending invite" on a real, visible group instead of a silent orphan.

Follow-up review passes then found the cleanup this fix shipped (`reap_orphaned_engine_groups`, for groups a *prior* release already orphaned) could itself delete a group still being created **or just joined**, that the pending-invite UI event never reached the frontend, and that resend used a possibly-consumed KeyPackage with no squad invite DM. Those are closed below.

## Design decisions

- **Persist-then-attempt-delivery.** `create_group` writes `mls_groups` metadata immediately after the engine commits, before any welcome is sent. A delivery failure now updates `pending_welcomes` on the existing row instead of the call returning `Err` and discarding the group.
- **Create/join-vs-reap lock.** `reap_orphaned_engine_groups` reads engine state and deletes anything with no matching metadata row — exactly what a group looks like between engine commit and metadata persist. `MLS_GROUPS_ENGINE_CREATE_LOCK` serializes that window for both `create_group` and `do_accept_mls_welcome` against every reap pass, then drops the lock before `STATE` so it cannot deadlock with paths that take STATE first.
- **Engine id from the welcome.** `do_accept_mls_welcome` persists `engine_group_id` from `welcome.mls_group_id` instead of falling back to `nostr_group_id` when `get_groups()` misses the new group. The reaper only matches engine hex ids, so that fallback would have left a live group unmatched and deletable.
- **Targeted upsert/delete over full-table replace.** `create_group`, `add_member_device`, and `leave_group` previously read the whole `mls_groups` table, mutated it in memory, and wrote the whole table back — a lost-update race where a concurrent call's freshly-persisted row could be silently dropped. They now upsert/delete only their own row. Pending-welcome clears are an element-scoped JSON update so concurrent resend/remove of different npubs cannot resurrect a just-cleared entry. Empty `engine_group_id` is guarded so the alias cannot rewrite every legacy row.
- **Resend no-op + latest KeyPackage.** The "Resend invite" action reuses the Restore (remove-then-re-add) path; two admins racing it could both run a full cycle. `add_member_device` takes an `is_resend` flag and, when set, re-checks `pending_welcomes` after acquiring the per-group lock. Resend skips the store-reset freshness gate (a pending member never reset) but still **fetches the latest KeyPackage**, because the one recorded at create may already have been consumed.
- **Squad invite DM after announcements resend.** Squad create withholds `squad_invite` from pending npubs so the invite card does not advertise a Welcome that never published. A successful **Resend invite** on the announcements group then sends that DM. Channel create still DMs pending npubs; a missing welcome makes accept a no-op, and the cached DM names the channel once the resend lands.

## Also fixed (review passes)

- `reap_orphaned_engine_groups` falls back to `group_id` when `engine_group_id` is empty (matching every other read site) and gets a 30s cooldown against relay-reconnect storms.
- `remove_member_device` clears `pending_welcomes` for the removed member and takes the same membership lock as add/resend.
- `mls_group_metadata` frontend listener reads nested `payload.metadata.group_id` (the backend emits `{ metadata: … }`); pending-invite notice now reaches channel creation, not just squad creation.
- Pending-welcome bookkeeping no longer fails `create_group` or a succeeded invite; a failed re-add keeps the npub pending so retry stays visible.
- All resend buttons disable while one is in flight; the members panel de-duplicates `renderNpubs`; unused `reportSkippedMembers` is gone.
- `PendingInvite` renamed to `UndeliveredInvite` (was one letter of confusion away from the pre-existing, differently-scoped `PendingMlsWelcome`).

## Testing

- `cargo test --lib`: 841 passed on the earlier commits in this PR, including `mls_groups` persist/re-persist/delete and the reap cooldown. This follow-up: `cargo test --lib mls_group_metadata_persistence_tests` — 9 passed (alias match, empty-id guard, element-scoped remove, sibling isolation).
- `pnpm test`: 2236 passed on the earlier commits, including resend no-op and pending-invite wiring. This follow-up: vitest `tauri-subscriptions`, `create-channel-flow`, `restore-mls-access`, `mls-group-members` — 64 passed (nested metadata event, post-resend squad DM, `pendingWelcomesByGroupId`).
- Earlier commits also ran `pnpm check` (0 errors), `pnpm lint` (clean), `pnpm check:tauri-commands` (no new orphans).

Fixes #321


daopunk (changes requested):
## Request changes

Persist-before-delivery is the right fix for #321, and the second commit’s create-vs-reap lock + targeted upserts are real. The recovery path this PR ships does not actually resend a welcome.

### Blocking: “Resend invite” runs Restore, so it cannot recover the bug

The UI comment says resend is “low risk, no revocation.” The backend does the opposite.

A pending member already has an engine leaf, so `add_member_device` classifies them as **Restore**: remove-then-re-add, admin-gated, and `resolve_fresh_keypackage` **rejects the KeyPackage this admin already recorded**.

That is fatal for this flow:

1. `create_group_chat` / `invite_member_to_group` both call `refresh_keypackages_for_contact` first, which writes the current KP event id into `mls_keypackages`.
2. Resend calls that refresh again, then `add_member_device(..., is_resend=true)`.
3. Restore looks up `recorded_ref` for that same `(npub, device_id)` and refuses any candidate with that id.
4. The invitee has not rotated keys — delivery failed, they did not reset — so after ~4×5s retries the call fails with “No post-reset keypackage observed yet… ask them to reopen the app.”

So the button that is supposed to finish #321 option A (“retry delivery on the existing group”) asks the invitee to republish a new KeyPackage, then revokes their leaf.

Even if the generation check were skipped:

- A failed re-add after the Restore remove already committed leaves them **out of the group** (`map_add_member_failure` / `restoration_remove_committed`).
- `get_mls_group_members` still lists them in `pending_welcomes`, but the panel only renders `{#each members}`. They disappear from the UI with a stale pending flag.

Issue A was: persist the group, **re-send the existing Welcome**. That means keep the epoch-1 welcome rumor and retry `gift_wrap_to`. Restore is the store-reset path; it is the wrong tool here.

### What is solid

- **Persist metadata + chat before delivery.** Engine commit and UI metadata are no longer one atomic “everyone got a welcome” step. A failed gift-wrap becomes `pending_welcomes` instead of `return Err` and an invisible MDK group.
- **`MLS_GROUPS_ENGINE_CREATE_LOCK`.** The reaper’s “engine group with no SQL row” predicate is exactly the create window. Serializing that is necessary.
- **Targeted `save_mls_group` / `delete_mls_group`.** Full-table `write_groups` was a lost-update class. Moving create / add / leave off it is correct.
- **Frontend wiring.** Pending toast on squad + channel create, `mls_group_metadata` subscriber, `UndeliveredInvite` vs `PendingMlsWelcome`, i18n en/es, `inviteMemberToGroup(..., isResend = false)` default so admit/invite callers stay valid.

### Other issues

**Welcome/member count mismatch still drops people silently.** Delivery only zips `min(welcomes, invited)`. Extra invited members keep a leaf and are neither delivered nor recorded in `pending_welcomes`. Pre-existing when all zipped sends succeeded; now create returns `Ok`, so it is easier to ship a member with no recovery handle. Mark leftover invited npubs pending.

**No test drives `create_group` with a failing `gift_wrap_to`.** #321 asked for that. What landed is a SQL round-trip that *mirrors* the persist/re-persist shape, plus a pure `pending_invites_from_delivery_outcomes` helper. That does not prove create no longer returns `Err`, does not prove the chat row is kept, and does not prove resend delivers.

**Docs still describe the old abort.** `docs/mls/INVITES_AND_MEMBERSHIP.md` and `docs/mls/TROUBLESHOOTING.md` do not mention persist-then-pending, resend, or the reaper. Code is now the source of truth; those pages are wrong.

**Squad invite DMs still go to pending members.** Navbar filters `skippedMembers` only. The invitee gets a `squad_invite` card for a Welcome that never published. Either skip pending npubs until resend succeeds, or say the invite is not accept-ready yet.

**Reaper: empty `known` deletes every engine group.** Intended for pre-fix orphans. `read_groups()` erroring is safe (`Err` → no delete); `Ok([])` with a live MDK store is not. Worth a guard (skip reap if SQL is empty and engine is not) or at least a loud log + dry-run count.

**`write_groups` / `save_mls_groups` still replace the whole table.** Hot paths moved off it; the helper remains and still documents “complete current list.” Any leftover or future caller can still drop a concurrently upserted row, and then the reaper will treat that group as an orphan.

**Second persist of `pending_welcomes` is unlocked.** Create drops `MLS_GROUPS_ENGINE_CREATE_LOCK` before the gift-wrap loop, then upserts from the in-memory `meta` clone. Unlikely on a group nobody else knows yet, but it can clobber a concurrent `add_member_device` on that same new group.

### Nits

- Module comments in `mls.rs` / `mls_orphan_reaper.rs` are review essays. Repo style is one short why-line; the design belongs in `docs/mls/`.
- `load_mls_groups` uses `serde_json::from_str(...).unwrap_or_default()` for `pending_welcomes` — corrupt JSON silently becomes “not pending.”
- Pending-clear in `add_member_device` matches `g.group_id == group_id` only; the resend no-op also matches `engine_group_id`. Same caller usually passes the wire id, but the two sites should use the same predicate.
- `hex` matching in the reaper is case-sensitive by test; harvest persist is case-insensitive. Fine while every writer uses `hex::encode` (lowercase).

### Suggested fix

Keep persist-before-delivery, the lock, and the SQL field. Change resend:

1. Persist (or reconstruct) the undelivered Welcome rumor on the create-failure path.
2. Resend = retry `gift_wrap_to` for that rumor. Do not go through Restore / `resolve_fresh_keypackage`.
3. Only after a successful gift-wrap, drop that npub from `pending_welcomes`.
4. Add one backend test: `create_group` + failing publish → group row exists, chat exists, `pending_welcomes` set, create returns `Ok`; resend publishes without `remove_members`.
5. Update `docs/mls/INVITES_AND_MEMBERSHIP.md`.

Until (2) is real, this PR stops minting new orphans but does not give the creator a working way to finish the invite.

daopunk (approved):
approve to not block

logicminds:
Worked the whole list. Addressed in 6cc3d051.

## Blocking — resend

> A pending member already has an engine leaf, so `add_member_device` classifies them as **Restore**: remove-then-re-add, admin-gated, and `resolve_fresh_keypackage` **rejects the KeyPackage this admin already recorded**.

Confirmed, and it's worse than "unless they rotated" — it could never succeed. `invite_member_to_group` calls `refresh_keypackages_for_contact` first, which writes the fetched event id into the index under a **synthetic `device_id` equal to that same event id**. So whatever the invitee publishes, `recorded_ref == candidate_id` on the following `add_member_device`, and `keypackage_generation_advanced` returns `false` for every candidate. Unconditionally broken.

Addressed differently, though: not by persisting and replaying the epoch-1 Welcome.

> Issue A was: persist the group, **re-send the existing Welcome**. That means keep the epoch-1 welcome rumor and retry `gift_wrap_to`. Restore is the store-reset path; it is the wrong tool here.

The leaf exists, so remove-then-re-add is *structurally* required — there is no path to a usable Welcome for that member that doesn't advance the epoch. What's actually wrong is narrower: the **store-reset KeyPackage-freshness gate** is being applied to a member who never reset. That gate exists because a reset member's recorded KeyPackage's private init key only lives in their archived store. A pending member's recorded KeyPackage is live, so gating their resend on rotation rejects it forever.

So the fix scopes the gate (and its "ask them to reopen the app so it republishes" error) to `action == Restore && !is_resend`. A resend resolves its KeyPackage like a first-time invite, and the remove-then-re-add mints a Welcome at the **current** epoch.

Replaying the stored epoch-1 rumor was the other option and I don't think it's the right one: an epoch-1 Welcome is unusable once the group has advanced, so it would work only in the narrow case where nothing else happened to the group, and it would add a new at-rest store of group secrets to carry rumors. The remove-then-re-add produces a Welcome that is correct by construction. Rationale is written up in `docs/mls/INVITES_AND_MEMBERSHIP.md` so it doesn't live only in this thread.

> A failed re-add after the Restore remove already committed leaves them **out of the group**… They disappear from the UI with a stale pending flag.

Fixed. The panel now renders `members ∪ pendingWelcomes` instead of `{#each members}`, so a member in that state stays visible with the badge and the resend button. A retry then classifies as `Invite` and self-heals. Also dropped the `resendInvite` comment claiming "low risk, no revocation" — you were right that it contradicted the backend.

## Other issues

> **Welcome/member count mismatch still drops people silently.** … Mark leftover invited npubs pending.

Fixed. Delivery goes through a `deliver_welcomes` seam that returns exactly one outcome per invited recipient; a recipient with no rumor at its index gets the no-welcome reason instead of falling off the end of a `min(len, len)` loop.

> **No test drives `create_group` with a failing `gift_wrap_to`.**

Addressed differently. `deliver_welcomes` takes the sender as a closure, and `deliver_welcomes_tests` drives it with failing and partially-failing senders composed through `pending_invites_from_delivery_outcomes`: all-fail, partial-fail, fewer-rumors-than-recipients (this one fails against the old zip behavior), and empty-rumors-with-recipients.

An end-to-end `create_group` test is not reachable from `cargo test`: `MlsService` resolves `TAURI_APP`, which is `OnceCell<AppHandle<Wry>>` and cannot be populated from `tauri::test::mock_app()`'s `MockRuntime`, and the path also needs a live nostr client and a `!Send` engine. Making it testable means threading `Runtime` through `MlsService`, which is a bigger change than this PR. The seam plus the existing SQL round-trip test cover the two things the fix depends on; a real create-with-failing-delivery run belongs in the Tauri debug harness.

> **Docs still describe the old abort.**

Fixed. `docs/mls/INVITES_AND_MEMBERSHIP.md` gets sections on undelivered invites/resend and on the reaper; `docs/mls/TROUBLESHOOTING.md`'s backend-behavior and user-guidance sections now separate a skipped member (needs to publish a KeyPackage) from a pending member (needs an admin resend).

> **Squad invite DMs still go to pending members.**

Fixed at all three sites (`Navbar.svelte`, and both flows in `squad-pair-create.ts`) — each builds one exclusion set of skipped ∪ pending. Chose "skip until resend succeeds" over "say it isn't accept-ready", since the card can only fail to accept. Covered by two new tests in `squad-pair-create.test.ts`.

> **Reaper: empty `known` deletes every engine group.**

Fixed. A `reap_would_wipe_every_group` predicate short-circuits the pass when metadata is empty and the engine is not, logging the count it declined to delete. Unit-tested.

> **`write_groups` / `save_mls_groups` still replace the whole table.**

Fixed by deleting them. After this PR they had zero production callers, so per the greenfield posture the helper and `replace_mls_groups_conn` are gone rather than documented — the lost-update class can't come back through a future caller.

> **Second persist of `pending_welcomes` is unlocked.**

Fixed, but by narrowing the write rather than widening the lock. All three mutation sites now call `db::set_mls_group_pending_welcomes`, which updates only `pending_welcomes` and `updated_at` and matches on `group_id OR engine_group_id`. A concurrent writer can no longer lose its other columns, and the lock stays off the network loop.

## Nits

> Module comments in `mls.rs` / `mls_orphan_reaper.rs` are review essays.

Fixed. The reaper's 18-line header is 5 lines, the blocks added to `mls.rs` are one-to-three-line why-lines, and the design moved to `docs/mls/INVITES_AND_MEMBERSHIP.md`. Also dropped the issue reference the no-tracker-IDs rule forbids.

> `load_mls_groups` uses `serde_json::from_str(...).unwrap_or_default()` for `pending_welcomes` — corrupt JSON silently becomes "not pending."

Fixed — logs the `group_id` and the parse error before degrading. Kept it degrading rather than failing so one corrupt row can't make the whole group list unreadable.

> Pending-clear in `add_member_device` matches `g.group_id == group_id` only

Fixed — folded into the change above; `set_mls_group_pending_welcomes` matches either alias, so the two sites now agree by construction.

> `hex` matching in the reaper is case-sensitive by test; harvest persist is case-insensitive.

Fixed rather than left alone. This path *deletes* groups, so a case mismatch is a destructive false positive; both sides now fold case, matching the harvest persist path. Returned ids keep engine-side casing for `hex::decode` and the logs.

---

Validation: `cargo test --lib` 846/846, `pnpm test` 2238/2238, `pnpm check` 0 errors, `pnpm lint` and `pnpm check:tauri-commands` clean.

PR #343, opened 2026-08-24
feat: speed up two-window sandbox demo

logicminds:
## Summary

Cuts dead time in the Host/Guest sandbox demo (`make dev-sandbox-pair`) without skipping MLS or changing production backup gating.

- Launch Host (`solo`) and Guest (`alice`) together with distinct autologin mnemonics, wait until both `sandbox-handle.json` files have an `npub`, then print MCP bridge ports.
- Write `npub` onto the sandbox handle after debug login/create-account so the driver can DM without a Settings round-trip (`#[cfg(debug_assertions)]`, no-op unless a sandbox handle exists).
- Under `PACTO_ALLOW_TEST_AUTH=1`, PIN save marks `backup_verified` the same way `dev_login` already does, and the UI reloads that flag so squad **Accept** is not a silent no-op.
- Catch-up welcome rows get an **Accept** button so Guest does not have to hunt the Host DM card.
- GIF disclosure can persist on a `_pending` key before `npub` is in the persistence context, then copy onto the npub-scoped key.
- Retry one profile SQLite open/migration on a transient disk I/O error during first-boot WAL create.

## Production impact

- Pair launcher, quiz skip, and handle `npub` writes stay behind debug / `PACTO_ALLOW_TEST_AUTH=1`.
- Catch-up Accept and the GIF pending-key fallback ship to real accounts. Backup reload after create-account only reads the existing flag.

## Test plan

- [ ] `make dev-sandbox-pair` starts two authed windows with different npubs and prints MCP ports
- [ ] Host DMs Guest → squad **Sandbox Demo** → Guest Accept (Catch up or Host DM) → `#welcome` → two chat lines → Guest GIF
- [ ] Without `PACTO_ALLOW_TEST_AUTH=1`, Accept still requires backup verification
- [ ] Release / non-sandbox login does not write `sandbox-handle.json`


daopunk (approved):
## Verdict: approve with comments

Mergeable for the stated Host/Guest demo goals. CI is green (typecheck, lint, frontend/backend/scripts/harness/symbol/landing). No critical/high blockers. One medium consistency issue on production Catch-up Accept; a few nits.

## What works well

- **Pair launcher** (`scripts/dev-sandbox-pair.sh`): distinct BIP-39 vectors (not Anvil fixture), mnemonic equality check, host-first compile, waits for handle then both `npub`s, Ctrl+C cleanup.
- **Handle `npub`**: `record_npub` behind `#[cfg(debug_assertions)]` + sandbox-root no-op — correct gating.
- **`await loadBackupVerified()`** after `createAccount`: fixes the race where Accept ran while the store was still `null`; test updated to expect `false`.
- **GIF pending key**: sensible pre-npub fallback with migration test.
- **SQLite disk I/O retry**: narrow classifier + single 50ms retry; message matching is unit-tested.
- **Demo agent prompt** aligns with Makefile/docs.

## Findings

### Medium — Catch-up Accept expands the backup-gate gap

Catch-up `accept()` calls `acceptOfferedWelcome()`, which never calls `requireBackupVerified()`. DM announcement Accept in `accept-invite.ts` does.

**Impact:** Production users with `backup_verified = false` can join MLS welcomes from Catch up while other invite Accept paths open the backup modal. Same gap already exists on `PendingWelcomeRequests.svelte`; this PR adds another production entry point (PR body acknowledges Catch-up Accept ships).

**Ask:** Gate inside `acceptOfferedWelcome()` (fixes both entry points) or at least in Catch-up `accept()`.

### Medium/Low — Accept always labeled Accept when welcome not matched

Catch-up shows Accept for every `entry.kind === 'welcome'`. If `$pendingMlsWelcomes` has no match yet, click falls through to `open()` with no toast — misleading during MLS latency. Prefer disable until `matchingWelcome` exists, or distinct copy.

### Low — `PACTO_ALLOW_TEST_AUTH` on `set_pkey` is env-only

`maybe_mark_backup_verified_for_test_auth` mirrors `dev_login` (env, not `debug_assertions`). Correct for sandbox Create Account; release still honors the env if someone exports it. Optional harden: also require sandbox root / `cfg(debug_assertions)`.

### Low — GIF `_pending` key not deleted after migrate; write-on-read

Migration copies to the npub key but leaves `_pending`. Shared-machine edge: account A accepts before npub is set → later account B can inherit via pending. Minor; clear `_pending` after migrate.

### Nit — unrelated file

`.agents/security-review-prompt.md` is a scheduled AppSec agent prompt with no tie to sandbox demo speed. Prefer drop or separate PR.

### Nit — duplicate JSDoc; temp logs; early `npub` on create_account

- Two consecutive `/** … */` above `offeredWelcomeFromPendingMls`.
- `mktemp` host/guest logs never removed.
- `record_npub` on `create_account` can set handle `npub` before PIN finishes — pair script may print “authenticated” while UI is still onboarding if autologin failed (primary mnemonic path is fine).

## Production vs debug gating

| Change | Production? |
|--------|-------------|
| `dev-sandbox-pair`, handle `npub` write | Debug/sandbox only |
| `PACTO_ALLOW_TEST_AUTH` backup mark on PIN save | Env-gated (same as `dev_login`) |
| Catch-up Accept, GIF pending key, `loadBackupVerified` await, SQLite retry | Yes |

## Non-blocking test gaps

- No Rust test that `set_pkey` + `PACTO_ALLOW_TEST_AUTH=1` sets `backup_verified`.
- No assertion that `acceptOfferedWelcome` requires backup verification.
- Shell pair script untested (acceptable; covered by PR manual checklist).

Approving as-is for demo velocity; please track the backup-gate consistency follow-up (ideally in `acceptOfferedWelcome`).

logicminds:
Addressing the review — fixed in 37617113.

> Catch-up `accept()` calls `acceptOfferedWelcome()`, which never calls `requireBackupVerified()` ... Gate inside `acceptOfferedWelcome()` (fixes both entry points) or at least in Catch-up `accept()`.

Fixed: `acceptOfferedWelcome()` now imports `requireBackupVerified` directly and throws before doing anything else if it fails. This closes the gap for both Catch-up Accept and `PendingWelcomeRequests.svelte` in one change since both call through this shared function; each already treats a throw as a failed accept via its existing try/catch. Added a test asserting the gate blocks the engine call.

> Catch-up shows Accept for every `entry.kind === 'welcome'`. If `$pendingMlsWelcomes` has no match yet, click falls through to `open()` with no toast ... Prefer disable until `matchingWelcome` exists.

Fixed: the Accept button is now `disabled={accepting || !matchingWelcome}`. Added `CatchUpEntry.test.ts` covering the disabled state.

> `maybe_mark_backup_verified_for_test_auth` mirrors `dev_login` (env, not `debug_assertions`) ... Optional harden: also require sandbox root / `cfg(debug_assertions)`.

Fixed: split the function into a `#[cfg(debug_assertions)]` real implementation and a `#[cfg(not(debug_assertions))]` no-op stub, matching `dev_login`'s gating — the `PACTO_ALLOW_TEST_AUTH` check no longer compiles into release binaries at all.

> Migration copies to the npub key but leaves `_pending` ... Minor; clear `_pending` after migrate.

Fixed: `isGifsDisclosureAccepted()` now clears the pending fallback key right after migrating it onto the npub-scoped key. Added a test.

> `.agents/security-review-prompt.md` is a scheduled AppSec agent prompt with no tie to sandbox demo speed. Prefer drop or separate PR.

Dropped. Confirmed no Makefile target, CI workflow, script, or doc references it.

> Two consecutive `/** … */` above `offeredWelcomeFromPendingMls`.

Fixed: removed the stray leftover doc comment (it actually described `offeredWelcomes` below it).

> `record_npub` on `create_account` can set handle `npub` before PIN finishes — pair script may print "authenticated" while UI is still onboarding if autologin failed.

Fixed: moved the sandbox-handle npub write out of `create_account` and into `set_pkey`'s pending-account branch, right after the DB/PIN save completes — matching how `dev_login` and the mnemonic-login path already time it.

> `mktemp` host/guest logs never removed.

Not addressing: `host_log`/`guest_log` paths are printed specifically so a Ctrl+C'd or crashed run can still be inspected afterward — that's the whole reason they're echoed to the user right before `wait`. Deleting them in `cleanup()` would remove the only way to debug a failed pair launch after the fact. They're standard OS-temp `mktemp` files, cleaned by the system's normal temp-file policy.

Validation: `pnpm test` (2495/2495), `pnpm check` (0 errors), `pnpm lint` (clean), `cargo test --lib` (925 passed, 2 pre-existing ignored).

PR #356, opened 2026-08-27
feat: add opt-in biometric unlock via touch id and windows hello

### LINKED ISSUE #179: Add biometric unlock (Touch ID / Windows Hello) via tauri-plugin-biometry, with a documented Linux gap
## Summary

Add an opt-in **biometric unlock** for the returning-user PIN screen — Touch ID on macOS and Windows Hello on Windows via [`tauri-plugin-biometry`](https://github.com/Choochmeque/tauri-plugin-biometry) (or an equivalent alternative) — so a returning user can unlock without typing their 6-digit PIN each launch. The manual PIN path stays as the mandatory fallback everywhere.

**Platform note (verified against the plugin's README):** `tauri-plugin-biometry` supports **macOS (Touch ID), Windows (Windows Hello), iOS, and Android** — it does **not** support Linux; there is no OS-native biometric prompt API on desktop Linux for it to wrap. This issue's title mentions Linux, but scope should ship macOS + Windows first and track Linux as a separate, explicitly-deferred follow-up (see "Out of scope" below) rather than promise a biometric prompt the chosen library can't deliver there.

## Current behavior

Unlock is PIN-only, no OS secure-storage or biometric integration exists in the codebase today (`src-tauri/Cargo.toml` has no `keyring`/`biometry`/`secure-storage` plugin).

- Frontend: `PinInput.svelte` collects a 6-digit PIN; `unlockWithPin(pin)` in `src/stores/auth.ts:293` calls `loadAndDecryptKey(pin)` (`src/lib/api/encryption.ts:56`), which invokes backend `get_pkey` (fetch the encrypted key) then `decrypt` (Argon2id-derived key + ChaCha20-Poly1305, see `src-tauri/src/crypto.rs`) to recover the Nostr private key, then completes login.
- Backend: `crypto::derive_key_from_salt` runs Argon2id (96 MiB memory, 6 iterations, `src-tauri/src/crypto.rs:29-51`) against the typed PIN to derive the 32-byte encryption key cached in `ENCRYPTION_KEY`. There is no path today that releases this key from anything other than a correct PIN.
- Every launch after account creation re-prompts for the PIN with no faster alternative.

## Proposed change

1. **Dependency:** Add `tauri-plugin-biometry` to `src-tauri/Cargo.toml` and `@choochmeque/tauri-plugin-biometry-api` to `package.json`, mirroring how other Tauri plugins are registered in `src-tauri/src/lib.rs`. Add the required capability entries (`biometry:default` plus scoped storage permissions for a single `domain` such as `io.pacto`) under `src-tauri/capabilities/`.
2. **Opt-in enrollment:** After a successful PIN unlock, if `checkStatus()` reports biometry available and the user hasn't already enrolled, offer "Enable biometric unlock" (Settings and/or a one-time post-unlock prompt). On accept, call `authenticate()` then store the material needed to reconstitute the decryption key — **not the raw PIN** — via the plugin's biometric-gated `setData({ domain: 'io.pacto', name: <npub>, data })`, scoped per-account like every other npub-keyed secret in this app.
3. **Unlock flow:** On the returning-user PIN screen (`Login.svelte`), when biometry is enrolled for the current/last account, show "Unlock with Touch ID" / "Unlock with Windows Hello" as the primary action ahead of `PinInput.svelte`, with an explicit "Use PIN instead" fallback always visible. On success, `authenticate()` → `getData()` → recover the key material → complete the same `unlockWithPin`-style login sequence (`activeTopNavTab`, `loadAccountState`, `currentUser`, etc.) without a typed PIN.
4. **Settings surface:** New toggle (e.g. "Unlock with biometrics") in the settings notifications/security area, persisted npub-scoped via `persistenceKey(prefix)` (`src/stores/persistence-context.ts`), independent of whether the OS-level secure storage still holds the enrolled secret — disabling the toggle should also call `removeData()` to actually revoke it, not just hide the button.
5. **Security review:** This changes how the encryption key can be released, in wallet/key-handling code the project already treats as alpha-grade (`docs/audits/README.md`). Explicitly scope what's stored via `setData` (prefer a device-bound wrapping secret over the raw decrypted key where the plugin's storage guarantees allow it) and get this path reviewed before it ships, not just tested.
6. **i18n:** Add strings to `src/lib/i18n/locales/en/` and `es/` for the new prompts/toggle/errors (unavailable biometry, lockout, cancelled).
7. **Tests:** Unit test the enrollment/unlock branching (biometry available vs. not, enrolled vs. not, `PinInput` fallback always reachable) with the plugin mocked, plus the settings toggle's persistence and its `removeData()` revoke path.

### Out of scope (this issue)

- Linux biometric/fingerprint unlock — no supported path in the chosen library; would need a separate investigation (e.g. a libsecret/polkit-backed prompt) and its own issue if pursued.
- Mobile (iOS/Android) enrollment UX — the plugin supports it, but this repo currently ships desktop only; revisit alongside any future mobile build.

## Acceptance criteria

- [ ] On macOS, a returning user can enable biometric unlock and subsequently unlock via Touch ID instead of typing the PIN.
- [ ] On Windows, the same works via Windows Hello.
- [ ] "Use PIN instead" is always available and works identically to today's flow, including when biometry is unavailable, not enrolled, cancelled, or fails.
- [ ] Disabling the setting revokes the stored secret (`removeData()`), not just the UI entry point.
- [ ] Nothing here weakens the existing PIN/Argon2id path: an attacker without the PIN and without a successful biometric prompt still cannot derive the encryption key.
- [ ] i18n strings added for `en` and `es`.
- [ ] Unit tests cover enrollment, unlock, fallback, and revoke branching.
- [ ] Linux and mobile are explicitly documented as out of scope for this issue, not silently unsupported.

logicminds:
> **Update:** the Settings screenshot below was captured before a later fix in this PR. The toggle now runs a silent capability probe before showing itself and stays **hidden on this project's current macOS signing** (ad-hoc, no Apple Developer Team ID — see `docs/build/macGuide.md`'s Troubleshooting section). It reappears automatically once the project has real Apple Developer ID signing; no code change needed then. See the PR comments for the full investigation.

## Summary

Returning users on macOS and Windows can now unlock Pacto with Touch ID or Windows Hello instead of typing their 6-digit PIN on every launch. The PIN path is unchanged and remains the mandatory fallback everywhere — Linux, devices without enrolled biometrics, a cancelled or failed prompt, and lockout all land back on the same `PinInput` screen that exists today.

The backend never releases the PIN or the raw private key to enable this: it exports the same Argon2id-derived session key a correct PIN already derives, and that key is what gets handed to the OS's biometric-gated secure storage (macOS Keychain / Windows Hello via `tauri-plugin-biometry`). A new Tauri capability scopes that plugin's storage commands to macOS and Windows only, enforcing the documented Linux gap at the permission layer in addition to the plugin's own unsupported-platform stub.

## Changes

- **Backend:** `export_encryption_key_material` and `unlock_with_biometric_key` Tauri commands; `tauri-plugin-biometry` registered and scoped via a new `capabilities/biometry.json` (macOS + Windows only).
- **Enrollment:** opt-in toggle in Settings → Security ("Unlock with Touch ID" / "Unlock with Windows Hello"), hidden entirely when biometry is unavailable. Disabling always calls `removeData()` to actually revoke the stored secret, not just hide the toggle.
- **Unlock:** the returning-user PIN screen offers the biometric prompt as the primary action when available and enrolled, with "Use PIN instead" always visible and functional.
- **Self-heal:** a stale/invalid biometric key (e.g. the OS secret was cleared behind the app's back) revokes the enrollment and falls back to PIN automatically instead of repeatedly offering a dead button.
- i18n strings added for `en` and `es`.

## Test plan

- `cargo check` (macOS/arm64, exercises the real Touch ID backend module) — clean.
- `cargo test parse_biometric_key_hex` — 5/5 passing (hex round-trip, whitespace, too-short, too-long, non-hex).
- `pnpm lint`, `pnpm check` (3160 files, 0 errors) — clean.
- `pnpm test` — 302 files / 2627 tests passing, including new coverage for enrollment/disable/revoke branching and the `unlockWithBiometrics` happy-path, self-heal, and gate-blocked cases.
- Manually verified against the seeded dev sandbox (`make dev-sandbox-seeded`) via the Tauri MCP bridge on this MacBook Pro (Touch ID enrolled): the Security toggle correctly detects biometry availability and renders "Unlock with Touch ID" (see screenshot). Actually completing enrollment requires a live Touch ID scan, which isn't something this session can trigger, so the enrolled/checked state and the Login-screen prompt were verified through the automated test suite instead.

## Screenshots

Settings → App settings → Security, on a Touch ID–enabled Mac, seeded dev sandbox:

Biometric unlock toggle in Settings > Security

## Out of scope

Linux biometric unlock and mobile (iOS/Android) enrollment UX are explicitly deferred, per the issue.

Closes #179


logicminds:
## Root cause of "Could not enable biometric unlock."

Reproduced by invoking `plugin:biometry|set_data` directly (bypassing the Touch ID prompt) via the Tauri MCP bridge:

```
[keychainError] - Error adding item to keychain: -34018
```

`-34018` is macOS's `errSecMissingEntitlement`. `tauri-plugin-biometry`'s macOS storage path uses `kSecUseDataProtectionKeychain` + `SecAccessControl` (biometric-gated Keychain items), which requires the app to be signed with a real **Apple Developer Team ID**.

Confirmed this is not a dev-sandbox artifact — I tested against **two** builds:

1. Raw `pnpm tauri dev` / `make dev-sandbox` binary (linker ad-hoc signed only, no bundle).
2. A properly bundled, ad-hoc-signed `.app` from `pnpm tauri build --debug --bundles app` (same `signingIdentity: "-"` this repo's `tauri.conf.json` already uses, and the same signing this repo's `release.yaml` uses today — it configures no Apple credentials).

Both fail identically. Ad-hoc signing (`codesign -s -`) never has a Team ID, so this isn't fixable by bundling/signing harder in code — it needs a genuine Apple Developer ID Application certificate wired into signing (and CI, for the shipped build to actually work for users). That's an infrastructure/business decision (Apple Developer Program enrollment, ~$99/yr) outside this PR's scope.

**Windows Hello is unaffected** — its storage backend is WebAuthn/Windows Credential Manager, not this Apple Keychain API.

### What's fixed here (`e5a6795`)

The toggle was silently swallowing the real error with no console output. Now it:
- logs the actual underlying error to the console, and
- detects this specific `keychainError` failure and shows a message naming the cause, instead of the generic "Could not enable biometric unlock."

### What's still open

Until this repo has a real Apple Developer ID certificate wired into macOS signing, the Settings toggle and the Login-screen prompt will always fail past the Touch ID prompt on macOS, in dev and in the current release build alike. Enrollment/unlock code itself is correct and covered by tests (verified via direct IPC calls, independent of the signing issue). Flagging for a decision on how to handle this before merge — options include: park macOS behind a documented known-limitation until signing is available, or treat this PR's macOS acceptance criterion as blocked on that infra work.

logicminds:
## Resolved: toggle now hides itself on this build (`72b1d76`)

Per discussion, rather than just documenting the limitation I added a proper fix: the Settings page now runs a silent capability probe (a real `setData`/`removeData` round trip with disposable data) before showing the Touch ID toggle. On this project's current macOS signing, the probe fails the same way `setData` did, so the toggle now stays **hidden** instead of offering a control that fails after a real Touch ID prompt.

The probe only runs for Touch ID — `setData` on Windows always shows an interactive OS prompt, so probing there would surprise users with a biometric dialog just for opening Settings. Windows Hello's toggle path is unaffected.

Verified live against the same unsigned dev-sandbox build that previously showed the toggle and failed: `document.body.textContent.includes('Unlock with Touch ID')` now returns `false`.

Documented the underlying constraint and how to lift it (real Apple Developer ID signing) in `docs/build/macGuide.md`'s Troubleshooting section, so this is self-explanatory the next time someone hits it and re-enables automatically once the project has real signing — no code change needed then, since the probe is a functional check, not a hardcoded platform gate.

Updated test coverage: `canPersistBiometricUnlockData` (probe success/failure/cleanup-failure) in `src/lib/api/biometry.test.ts`.

logicminds:
Reopening to re-trigger CI checks after a squash push that GitHub Actions didn't pick up.

